PR #1, opened 2026-01-27
Add Flyway database plugin for Aurora DSQL

amaksimo:
## Summary

Adds Java/Flyway adapter to the monorepo - the first Java adapter.

### Implementation (5 classes)
- `AuroraDSQLDatabaseType` - URL detection, driver class, priority
- `AuroraDSQLDatabase` - DDL transaction handling, schema history table
- `AuroraDSQLConnection` - SET ROLE bypass, advisory lock bypass
- `AuroraDSQLSchema` - Custom doClean() that drops views before tables
- `AuroraDSQLTable` - FOR UPDATE bypass

### Tests
- 25 unit tests
- 11 integration tests against live DSQL cluster

### CI/CD
- `java-flyway-ci.yml` - Build, unit test, integration test with DSQL cluster
- `java-flyway-release.yml` - Release workflow (Maven Central publishing TBD)

## DSQL Compatibility

The plugin handles these DSQL-specific behaviors:
- SET ROLE not supported (DSQL uses IAM authentication)
- Advisory locks not supported (DSQL uses OCC)
- One DDL statement per transaction
- FOR UPDATE requires equality predicates on key
- Views must be dropped before tables in clean

## Testing

Unit tests pass locally. Integration tests verified against DSQL cluster `aftp56tv4kdk4ld7qyakksrlu4.dsql.us-east-1.on.aws`.

## Next Steps

- Maven Central publishing configuration (GPG keys, Sonatype credentials, pom.xml metadata)


.github/workflows/java-flyway-ci.yml
```diff
@@ -0,0 +1,90 @@
...
+    working-directory: java/flyway
+
+jobs:
+  build:
+    name: Build and Unit Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up JDK 17
```
Benjscho: Should we be using JDK 21?

java/flyway/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+- Verify IAM permissions include `dsql:DbConnectAdmin`
+- Check AWS credentials are configured (environment variables, IAM role, or credentials file)
+- Ensure the AWS region is correctly set
+
+### "change conflicts with another transaction, please retry" (OC000)
+
+This is an optimistic concurrency control (OCC) conflict. Aurora DSQL uses OCC instead of locks, and conflicts are detected at commit time. This error is rare during migrations since they typically run single-threaded.
+
+If you encounter this error, simply re-run `flyway migrate`. Flyway migrations are idempotent - already-applied migrations will be skipped.
```
Benjscho: How does this work if a migration fails partway through and then becomes unrecoverable?

java/flyway/src/test/java/software/amazon/dsql/flyway/AuroraDSQLFlywayIntegrationTest.java
pkale, line 1: We should add a basic OCC test, even if its testing with basic retry. This seems important to me given that OCC is a fundamental dsql characteristic

java/flyway/src/main/java/software/amazon/dsql/flyway/AuroraDSQLConnection.java
```diff
@@ -0,0 +1,61 @@
...
+        super(database, connection);
+    }
+
+    /**
+     * Skips SET ROLE restoration - DSQL uses IAM authentication where role is fixed.
+     */
+    @Override
+    protected void doRestoreOriginalState() throws SQLException {
+        LOG.fine("Skipping SET ROLE restoration (not supported by Aurora DSQL)");
+    }
```
Benjscho, line 37: What does this do in flyway? Is it appropriate to skip instead of erring?

java/flyway/README.md
pkale: Imo I think this guidance depends on the usecase and our expected audience. Are we expecting customers to use this just in ci/cd pipelines or in production. If we are expecting them to use it in production, we might want to consider adding retry behavior instead of manual intervention required. 

Ex: 
```
For CI/CD pipelines and manual migrations, simply re-run `flyway migrate` 
if you encounter an OC000 error.

For production applications that run migrations automatically, consider implementing retry logic.
```

java/flyway/src/main/java/software/amazon/dsql/flyway/AuroraDSQLDatabaseType.java
```diff
@@ -0,0 +1,100 @@
...
+ * behaviors. Discovered via Java SPI when the JAR is on the classpath.</p>
+ *
+ * <p>The Aurora DSQL JDBC Connector transforms URLs from {@code jdbc:aws-dsql:postgresql://}
+ * to {@code jdbc:postgresql://} internally. We detect DSQL by both URL prefix and endpoint
+ * pattern (*.dsql.*.on.aws).</p>
+ */
+public class AuroraDSQLDatabaseType extends PostgreSQLDatabaseType {
+
+    // Pattern to detect DSQL endpoints: *.dsql.<region>.on.aws
+    private static final String DSQL_ENDPOINT_PATTERN = ".dsql.";
```
Benjscho: Can this support private link endpoints (e.g., `dsql-fnh4`)?

java/flyway/README.md
amaksimo: What you are describing is if the migration fails part way and then we have something committed and need to fix the schema right? 

User can repair schema via flyway repair command, I can fix the readme to explain that. However it is extremely unlikely to get OCC error here, given this is a migration tool running schema, OCC error would happen once the user interacts with the db after migrations are done (unrelated to this). Wondering if I should just remove this section.

Flyway is single threaded and does not write concurrently, so I think we can safely remove this.

.github/workflows/java-flyway-ci.yml
amaksimo: Agree, will change.

java/flyway/README.md
amaksimo: I will remove this section, see the reply to Ben. It is not a realistic concern, OCC error would come after user has run flyway migrations and starts using the db (which is unrelated to this commit).

java/flyway/src/main/java/software/amazon/dsql/flyway/AuroraDSQLConnection.java
amaksimo: Flyway uses `SET ROLE` to restore the original database role after operations. 

DSQL doesn't support SET ROLE because roles are fixed via IAM (so we do not need to worry about it).

If we errored, Flyway would fail on every operation.

java/flyway/src/main/java/software/amazon/dsql/flyway/AuroraDSQLDatabaseType.java
amaksimo: Great comment! I missed that :D will fix.

java/flyway/src/test/java/software/amazon/dsql/flyway/AuroraDSQLFlywayIntegrationTest.java
amaksimo: Replied in previous comment, we do not need to worry about this here (single threaded single migrations for flyway).

java/flyway/src/main/java/software/amazon/dsql/flyway/AuroraDSQLDatabaseType.java
```diff
@@ -21,13 +21,12 @@
...
  * <p>The Aurora DSQL JDBC Connector transforms URLs from {@code jdbc:aws-dsql:postgresql://}
  * to {@code jdbc:postgresql://} internally. We detect DSQL by both URL prefix and endpoint
- * pattern (*.dsql.*.on.aws).</p>
+ * pattern (*.dsql.*). This supports both public endpoints and PrivateLink endpoints.</p>
  */
 public class AuroraDSQLDatabaseType extends PostgreSQLDatabaseType {
 
-    // Pattern to detect DSQL endpoints: *.dsql.<region>.on.aws
+    // Pattern to detect DSQL endpoints (supports both public and PrivateLink endpoints)
     private static final String DSQL_ENDPOINT_PATTERN = ".dsql.";
```
Benjscho: This still doesn't look like it supports PL endpoints, `<cluster-id>.dsql-fnh4.us-east-1.on.aws` would fail

amaksimo: Fixed

Benjscho (approved):

PR #6, opened 2026-01-28
Remove unused dsql.region system property from Flyway build

danielfrankcom:
Removes the unused `dsql.region` system property from the Flyway adapter's Gradle build configuration.

The property was being set from `AWS_REGION` but never read anywhere in the codebase. Only `dsql.cluster.endpoint` is actually used by the integration tests.


danielfrankcom:
Removing since this was moved to the [github.com/awslabs/aurora-dsql-tools](https://github.com/awslabs/aurora-dsql-tools) repo

PR #9, opened 2026-01-28
Add missing gitleaks-scan workflow

danielfrankcom:
This PR ports the `gitleaks-scan.yml` action from the original repos to perform secret scanning.


danielfrankcom:
I think this should be invalidated by #13 since I added it as a pre-commit check, so closing this.

PR #11, opened 2026-01-28
Format Java files

danielfrankcom:
This PR applies auto-formatting to Java files, in preparation for an upcoming change to the pre-commit checks.

We were previously enforcing formatting for the other subprojects, so I am combining those checks to run at the root level. I've split this change out to make the final diff smaller and isolate formating-only changes. There should be no functional difference before/after.

### DIFF
--- a/java/flyway/src/main/java/software/amazon/dsql/flyway/AuroraDSQLConnection.java
+++ b/java/flyway/src/main/java/software/amazon/dsql/flyway/AuroraDSQLConnection.java
@@ -4,58 +4,55 @@
  */
 package software.amazon.dsql.flyway;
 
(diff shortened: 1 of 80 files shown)


danielfrankcom:
Looks like there is a competing format somewhere in the build, will try to figure out which it uses and more that to the pre-commit checks if possible.

danielfrankcom:
There was a formatting plugin used by only the Hibernate example, but it didn't apply to the dialect code. I've removed it since we are now formatting at the top level so it is unnecessary for the example to have a dedicated plugin. Conflict should be resolved.

amaksimo (approved):

PR #20, opened 2026-01-29
Update changelog files as part of release

danielfrankcom:
This PR adds a workflow step to update the `CHANGELOG.md` files as part of the release, based on the release notes in the GitHub releases. This provides a forge-agnostic change history for the repo.

This was previously set up for the Django and Tortoise ORM repos, but not for the others. I've added the workflow for all subprojects, and manually synced the `CHANGELOG.md` files to reflect the current state of the GitHub releases.

I cloned the releases from the original repositories to provide easier access for users, and to make generating these files easier going forward. It is expected that the release notes still point to commits from the archived repo, since it will be left read-only, and those pull requests don't exist anywhere else.


amaksimo (approved):
Thanks! I guess we will want to add this to tools and connectors as well :)

PR #21, opened 2026-01-29
Wait for CI before releasing

danielfrankcom:
This PR adds a `wait-for-ci` job to all release workflows to ensure CI passes before publishing packages.

Previously, pushing a release tag would immediately trigger the build and publish steps, even if CI was failing on that commit. This could result in broken packages being published to PyPI/Maven Central.

Now the release workflows wait for CI to complete successfully before proceeding. This was previously being done for some of the isolated repos, but is now applied to all workflows in the monorepo.


amaksimo (approved):
lgtm, thanks

PR #80, opened 2026-01-31
Update top-level README to reflect all packages

danielfrankcom:
This PR makes a few changes to the top-level `README.md` to align with the packages contained in this repo:
- Adds a license column to the table to make the Hibernate dialect license discrepancy clearer
- Adds a Node.js table for the Prisma tooling subproject

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,19 +6,25 @@ This monorepo contains ORM adapters for [Amazon Aurora DSQL](https://aws.amazon.
 
 ## Available Adapters
 
+### Node.js
+
+| Package | Description | npm | License(s) |
+|---------|-------------|-----|------------|
+| [@aws/aurora-dsql-prisma-tools](./node/prisma/) | CLI tools for using Prisma with Aurora DSQL | [![npm](https://img.shields.io/npm/v/@aws/aurora-dsql-prisma-tools)](https://www.npmjs.com/package/@aws/aurora-dsql-prisma-tools) | ![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg) |
+
 ### Python
 
-| Package | Description | PyPI |
-|---------|-------------|------|
-| [aurora-dsql-django](./python/django/) | Django database backend for Aurora DSQL | [![PyPI](https://img.shields.io/pypi/v/aurora-dsql-django)](https://pypi.org/project/aurora-dsql-django/) |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,19 +6,25 @@ This monorepo contains ORM adapters for [Amazon Aurora DSQL](https://aws.amazon.
...
 ### Python
 
-| Package | Description | PyPI |
-|---------|-------------|------|
-| [aurora-dsql-django](./python/django/) | Django database backend for Aurora DSQL | [![PyPI](https://img.shields.io/pypi/v/aurora-dsql-django)](https://pypi.org/project/aurora-dsql-django/) |
-| [aurora-dsql-sqlalchemy](./python/sqlalchemy/) | SQLAlchemy dialect for Aurora DSQL | [![PyPI](https://img.shields.io/pypi/v/aurora-dsql-sqlalchemy)](https://pypi.org/project/aurora-dsql-sqlalchemy/) |
-| [aurora-dsql-tortoise-orm](./python/tortoise-orm/) | Tortoise ORM adapter for Aurora DSQL | [![PyPI](https://img.shields.io/pypi/v/aurora-dsql-tortoise-orm)](https://pypi.org/project/aurora-dsql-tortoise-orm/) |
+| Package | Description | PyPI | License(s) |
+|---------|-------------|------|------------|
+| [aurora-dsql-django](./python/django/) | Django database backend for Aurora DSQL | [![PyPI](https://img.shields.io/pypi/v/aurora-dsql-django)](https://pypi.org/project/aurora-dsql-django/) | ![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg) |
```
wcmjunior, line 19: Heads up:

In case this README is ever embedded, these relative links will break.

wcmjunior (approved):
Thanks for this change @danielfrankcom ! I added a minor comment below related to the relative references, but don't consider it as a deal breaker: more like loud thoughts I had while reviewing it. :-)

README.md
danielfrankcom: I'm not sure if you have a specific embed location you're thinking about, but I'm not sure we would publish this one to any package managers given its too broad compared to the `README.md` files within each subproject.

On a more general note, my preference is to use local paths in these files as they work for a local clone of the repo in an IDE, and we can transform the paths to be portable for package managers like we do [here](https://github.com/awslabs/aurora-dsql-orms/blob/85b2bf5f72e22742ddbdc5301e88df19456c4b91/python/tortoise-orm/pyproject.toml#L68) for the Python projects

PR #106, opened 2026-02-03
Simplify node-prisma release workflow

amaksimo:
## Summary

- Remove artifact-based publishing in favor of build+publish in same job
- Aligns with nodejs-connector release pattern
- Uses NPM_TOKEN for first publish (OIDC can't create new packages)

## Changes

- Removed separate `build` job with artifact upload/download
- Build and publish now happen in `publish-to-npm` job
- Added `--provenance` flag for npm provenance tracking
- Uses `NODE_AUTH_TOKEN` from `NPM_TOKEN` secret

## Test plan

- [ ] Merge PR
- [ ] Delete existing tag and re-create to trigger release
- [ ] Verify npm publish succeeds

### DIFF
--- a/.github/workflows/node-prisma-release.yml
+++ b/.github/workflows/node-prisma-release.yml
@@ -23,10 +23,15 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           wait-interval: 30
 
-  build:
-    name: Build distribution
+  publish-to-npm:
+    name: Publish to npm
     needs: wait-for-ci
     runs-on: ubuntu-latest
+    environment:
+      name: npm
+    permissions:
+      id-token: write # required for npm trusted publishing (OIDC)
+
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/node-prisma-release.yml
```diff
@@ -36,47 +41,18 @@ jobs:
...
-      - name: Set up Node
-        uses: actions/setup-node@v6
-        with:
-          node-version: "22.x"
-          registry-url: "https://registry.npmjs.org"
-
       - name: Publish to npm
-        working-directory: package/node/prisma
-        run: npm publish
+        run: npm publish --provenance
```
danielfrankcom, line 53: Isn't this flag unneeded since it's in the `package.json`? IIRC

danielfrankcom (approved):

.github/workflows/node-prisma-release.yml
amaksimo: I am not super sure, but we will revert this after the initial publish.

PR #311, opened 2026-02-28
Tortoise 1.0 fixes and don't require both asyncpg + psycopg to be installed.

hfern:
This PR fixes two issues preventing using the DSQL Tortoise ORM integration:

1. The module used to require having both `asyncpg` and `psycopg` installed. This is because importing the `aurora_dsql_tortoise` root module (which occurs when Tortoise uses importlib to import eg `aurora_dsql_tortoise.asyncpg`) ends up importing each submodule's register functions, which transitively imports the underlying backend module.

    I added some deferred imports and added a test so that if you are planning to use `$backend`, you can import all of the core parts of this module without them importing the other backends. The ImportError errors at least give you a nice stack trace and error description in the test.

    Truthfully, we should probably use tox to test a couple of "expected to be good" environments instead of patching the import system in a unit test. But we may add that later.

3. Tortoise1 made a bunch of changes compared to <1. One of the things they changed was the template arguments so that table_name is escaped _before_ being provided to the template. The net effect is that if you use this DSQL module with the newer Tortoise v1, this module tries doing:

    `CREATE INDEX ASYNC IF NOT EXISTS ... ON ""test_parent"" (...);`

    Note that double quoting around the table name, which is a syntax error. I did not want to drop support for tortoise<1, so I just support both heuristically in this PR. Hopefully when you guys decide to drop support for <1, you can just remove that branch.

    Another thing they changed is the inbuilt migrations mechanism. Due to them changing how the PK field SQL is calculated, this package can't inject the `BIGSERIAL => GENERATED BY DEFAULT AS IDENTITY` shim. That's one problem here. The other is that the migration table tortoise ships uses `SERIAL` anyways (not `BIGSERIAL`) so the inbuilt migration mechanism is somewhat moot anyways.

### DIFF
--- a/python/tortoise-orm/aurora_dsql_tortoise/__init__.py
+++ b/python/tortoise-orm/aurora_dsql_tortoise/__init__.py
@@ -4,8 +4,18 @@
 __all__ = ["register_backends", "register_asyncpg", "register_psycopg"]
 
 from aurora_dsql_tortoise._version import __version__ as __version__
-from aurora_dsql_tortoise.asyncpg import register_backend as register_asyncpg
-from aurora_dsql_tortoise.psycopg import register_backend as register_psycopg
+
+
+def register_asyncpg():
+    # Defer import to avoid depending on asyncpg in __init__
+    from aurora_dsql_tortoise.asyncpg import register_backend as _register_asyncpg
+    _register_asyncpg()
+
+
+def register_psycopg():
+    # Defer import to avoid depending on psycopg in __init__
+    from aurora_dsql_tortoise.psycopg import register_backend as _register_psycopg
+    _register_psycopg()
 
 
 def register_backends():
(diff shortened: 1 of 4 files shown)


Benjscho:
Thanks for the contribution! We'll take a look at the issues with the CI workflow, but this looks good to me

amaksimo:
> Thanks for the contribution! We'll take a look at the issues with the CI workflow, but this looks good to me

Taking a look at it, it is because it is a fork PR. We do not pass IAM secrets required to run the CI. I will run them manually and merge. Thanks!

* [Link to manual run](https://github.com/awslabs/aurora-dsql-orms/actions/runs/22637973397)

amaksimo (approved):

amaksimo (approved):
Verified via workflow_dispatch CI run - all tortoise tests pass (unit + integration on Python 3.10-3.13). Other ecosystem failures were GitHub infrastructure outages, not code issues.

hfern:
Thank you guys for the cool service, the library support, and merging through this PR :clap:

PR #339, opened 2026-03-19
Update Prisma CLI for DSQL sequence support

amaksimo:
## Summary

DSQL now supports sequences via `GENERATED BY DEFAULT AS IDENTITY`. This PR updates the Prisma CLI tooling to reflect that:

**Validator (`validate.ts`):**
- `autoincrement()` is now valid but emits a **warning** about hot keys (sequential IDs create uneven distribution across DSQL's distributed nodes)
- `@db.Serial`/`@db.SmallSerial`/`@db.BigSerial` still **error** — DSQL doesn't support SERIAL syntax, users should use `autoincrement()` instead
- Removed the old `Int @id` and `BigInt @id` warnings

**Transformer (`transform.ts`):**
- Added `SERIAL` → `INTEGER GENERATED BY DEFAULT AS IDENTITY` conversion
- Added `BIGSERIAL` → `BIGINT GENERATED BY DEFAULT AS IDENTITY` conversion
- Added `SMALLSERIAL` → `SMALLINT GENERATED BY DEFAULT AS IDENTITY` conversion
- This closes the gap where Prisma desugars `autoincrement()` to `SERIAL` in migration SQL

**The full pipeline is now:**
`@default(autoincrement())` in schema → Prisma generates `SERIAL` in SQL → transformer rewrites to `GENERATED BY DEFAULT AS IDENTITY` → DSQL handles it natively.

**Other changes:**
- Updated README validator/transformer tables and schema requirements docs
- Renamed `checkUnsupportedFeatures` → `checkFulltextIndex` + `checkSerialTypes` for clarity
- Cleaned up CLI help text
- Updated and consolidated tests (63 passing)

## Test plan
- [x] All 63 tests pass locally (`npm test`)
- [x] Prettier formatting clean (`prettier --check .`)
- [ ] CI passes


node/prisma/src/cli/transform.ts
```diff
@@ -370,6 +380,39 @@ function extractComment(statement: string): { comment: string; sql: string } {
...
+    count++;
+    return "SMALLINT GENERATED BY DEFAULT AS IDENTITY";
+  });
+
+  // SERIAL → INTEGER GENERATED BY DEFAULT AS IDENTITY
+  // Must come after BIGSERIAL/SMALLSERIAL to avoid partial matches
+  result = result.replace(/\bSERIAL\b/gi, () => {
+    count++;
+    return "INTEGER GENERATED BY DEFAULT AS IDENTITY";
+  });
```
Benjscho: We only support BIGINT

node/prisma/src/cli/transform.ts
```diff
@@ -370,6 +380,39 @@ function extractComment(statement: string): { comment: string; sql: string } {
...
+ * We convert to:    "id" INTEGER GENERATED BY DEFAULT AS IDENTITY NOT NULL
+ */
+function convertSerialTypes(sql: string): { sql: string; count: number } {
+  let count = 0;
+  let result = sql;
+
+  // BIGSERIAL → BIGINT GENERATED BY DEFAULT AS IDENTITY
+  result = result.replace(/\bBIGSERIAL\b/gi, () => {
+    count++;
+    return "BIGINT GENERATED BY DEFAULT AS IDENTITY";
```
Benjscho: It requires a cache value after identity: https://docs.aws.amazon.com/aurora-dsql/latest/userguide/sequences-identity-columns-overview.html

node/prisma/src/cli/transform.ts
amaksimo: Fixed

node/prisma/src/cli/transform.ts
amaksimo: Fixed, also added configuration for user.

node/prisma/src/cli/validate.ts
```diff
@@ -7,6 +7,11 @@
  * https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-unsupported-features.html
  *
  * If DSQL adds support for any of these features, update this validator and the README table.
+ *
+ * NOTE: As of 2025, DSQL supports sequences via GENERATED BY DEFAULT AS IDENTITY.
+ * autoincrement() is valid but triggers a warning (sequential IDs can create hot keys).
+ * The previous errors for autoincrement(), Int @id, and BigInt @id have been removed.
+ * SERIAL/BIGSERIAL/SMALLSERIAL syntax is still not supported by DSQL.
```
Benjscho, line 14: I don't think this comment needs to be here, this is more of a changelog that the commit will contain

node/prisma/test/transform.test.ts
```diff
@@ -147,92 +312,366 @@ ALTER TABLE "post" ADD CONSTRAINT "post_authorId_fkey" FOREIGN KEY ("authorId")
...
+     * Executes transformed SQL on DSQL and verifies an insert with
+     * auto-generated identity works. No-op when not connected.
+     */
+    async function verifyIdentityOnDsql(
+      transformedSql: string,
+      table: string,
+      insertColumn: string,
+      insertValue: string,
+    ): Promise<void> {
+      if (!dsql) return;
```
Benjscho, line 341: Why are we no-oping tests if we can't connect?

anwesham-lab:
# PR #339 Review — Update Prisma CLI for DSQL Sequence Support

**Repository:** `awslabs/aurora-dsql-orms`
**Branch:** `remove-prisma-sequence-checks` → `main`
**Author:** @amaksimo | **Commit:** `fdeea79`
**Stats:** +1,540 / -609 across 10 files (~1,400 lines are `package-lock.json`; actual logic is ~200-300 lines)

---

## Executive Summary

This PR updates the Prisma CLI tooling to support DSQL's new `GENERATED BY DEFAULT AS IDENTITY` feature. The core transformer and validator logic is solid, the BIGINT-only mapping is correct per [DSQL docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/create-sequence-syntax-support.html), and test coverage is strong (63 tests). Several substantive issues need attention — most notably silent test skipping, an `--identity-cache` edge case that produces invalid SQL, and fragile substring matching in the validator.

---

## Issues

### :red_circle: Silent No-Op Integration Tests

**File:** `node/prisma/test/transform.test.ts`

**15+ integration tests** silently pass when no DSQL cluster is available. The `if (!dsql) return` pattern in helper functions makes tests appear green when they actually ran nothing.

Three separate patterns contribute:

**`executeOnDsql` helper (~line 110):**
```typescript
async function executeOnDsql(transformedSql: string, tables: string[]): Promise<void> {
  if (!dsql) return;  // silent skip
```
Affects tests across `basic transformations`, `CREATE INDEX`, `foreign key removal`, `already wrapped statements`, `real-world Prisma output`, and `edge cases` describes.

**`verifyIdentityOnDsql` helper (~line 340):**
```typescript
async function verifyIdentityOnDsql(...): Promise<void> {
  if (!dsql) return;  // silent skip
```
Affects all SERIAL conversion tests — the INSERT RETURNING verification is entirely skipped.

**Inline `if (dsql) { ... }` blocks** in individual tests (e.g., "all serial variants map to BIGINT", "wraps DROP TABLE/INDEX", "preserves non-FK ALTER TABLE").

**Recommendation:** Use `describe.skip` / `test.skip` with a clear message (e.g., `"Skipping: CLUSTER_ENDPOINT not set"`) so CI output transparently shows what ran vs. what was skipped. This aligns with [Benjscho's review comment](https://github.com/awslabs/aurora-dsql-orms/pull/339#discussion_r2982965024).

---

### :red_circle: `--identity-cache` Accepts Scientific Notation → Invalid SQL

**File:** `node/prisma/src/cli/index.ts` (CLI parsing) + `node/prisma/src/cli/transform.ts` (`validateIdentityCache`)

`--identity-cache 1e100` passes all validation and produces invalid SQL:

1. CLI: `Number("1e100")` → `1e+100` (not `NaN`, passes `isNaN` check)
2. `validateIdentityCache`: `Number.isInteger(1e+100)` → `true` in JS (large floats pass this check)
3. `1e+100 >= 65536` → `true`
4. SQL output: `CACHE 1e+100` — **invalid SQL syntax**

Similarly, `--identity-cache Infinity` passes the CLI `isNaN` check (though `validateIdentityCache` catches it via `Number.isInteger`).

**Fix:** Add `Number.isSafeInteger(cache)` check or cap at a reasonable max, and add a test for scientific notation input.

---

### :red_circle: PR Description Contradicts Implementation

The PR description states:
> - SERIAL → INTEGER GENERATED BY DEFAULT AS IDENTITY
> - BIGSERIAL → BIGINT GENERATED BY DEFAULT AS IDENTITY
> - SMALLSERIAL → SMALLINT GENERATED BY DEFAULT AS IDENTITY

But the implementation maps **all** variants to `BIGINT GENERATED BY DEFAULT AS IDENTITY`. The code is correct — [DSQL's CREATE SEQUENCE](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/create-sequence-syntax-support.html) explicitly states `where data_type is BIGINT`, and [reviewer Benjscho confirmed](https://github.com/awslabs/aurora-dsql-orms/pull/339#discussion_r2968114044) "We only support BIGINT."

**Action:** Update the PR description to match reality:
> Added SERIAL/BIGSERIAL/SMALLSERIAL → `BIGINT GENERATED BY DEFAULT AS IDENTITY (CACHE n)` conversion (DSQL only supports BIGINT for sequences/identity columns)

---

### :yellow_circle: Fragile Substring Matching in `checkSerialTypes`

**File:** `node/prisma/src/cli/validate.ts` — `checkSerialTypes` function (~line 117)

```typescript
if (line.includes("@db.SmallSerial") || line.includes("@db.BigSerial")) {
  // error: Serial types not supported
} else if (line.includes("@db.Serial")) {
  // error: @db.Serial not supported
}
```

`line.includes("@db.Serial")` also matches `@db.SmallSerial` and `@db.BigSerial` since both contain the substring `@db.Serial`. This is currently masked by the `else if` ordering — but if the first branch is removed or reordered, the second branch produces false positives. A regex with exact matching (e.g., `/@db\.Serial\b/` or checking for `@db.Serial` followed by a non-alpha char) would be more robust.

---

### :yellow_circle: Changelog-Style Comment in Source Code

**File:** `node/prisma/src/cli/validate.ts` (lines ~10-14)

```typescript
 * NOTE: As of 2025, DSQL supports sequences via GENERATED BY DEFAULT AS IDENTITY.
 * autoincrement() is valid but triggers a warning (sequential IDs can create hot keys).
 * The previous errors for autoincrement(), Int @id, and BigInt @id have been removed.
```

Per [Benjscho's review](https://github.com/awslabs/aurora-dsql-orms/pull/339#discussion_r2982948563), this reads like a changelog ("The previous errors ... have been removed") rather than code documentation. Rewrite to describe the *current* validation behavior, not the history of what changed.

---

### :yellow_circle: Breaking Behavioral Change Not Documented

This PR changes user-visible behavior that should be explicitly called out in the description:

- **`autoincrement()` was previously an error, now a warning.** Users relying on the validator to block `autoincrement()` schemas in CI will see different behavior.
- **Two warnings removed:** `Int @id without autoincrement` and `BigInt @id` warnings no longer emitted.
- **Error messages changed:** `@db.Serial` errors now reference "IDENTITY" instead of "sequences."

**Recommendation:** Add a "Behavioral Changes" section to the PR description so downstream users are aware.

---

### :yellow_circle: `afterAll` Cleanup Swallows All Errors

**File:** `node/prisma/test/transform.test.ts` (~line 100)

```typescript
afterAll(async () => {
  if (dsql) {
    for (const table of [...tablesToCleanup].reverse()) {
      await queryWithRetry(dsql, `DROP TABLE IF EXISTS "${table}"`).catch(() => {});
    }
    await dsql.end();
  }
});
```

`.catch(() => {})` silently swallows ALL errors — not just expected ones. If cleanup fails for a meaningful reason (connection dropped, permission issues), orphaned test tables persist in the DSQL cluster with no diagnostic output. At minimum, log the error: `.catch((e) => console.warn("Cleanup failed:", e.message))`.

---

### :yellow_circle: SERIAL in SQL Comments Could Be Falsely Converted

**File:** `node/prisma/src/cli/transform.ts` — `convertSerialTypes`

The `\bSERIAL\b` regex operates on full SQL text including comments. A comment like `-- This was originally a SERIAL column` would be rewritten to `-- This was originally a BIGINT GENERATED BY DEFAULT AS IDENTITY (CACHE 1) column`.

In practice, Prisma's `migrate diff` uses only leading `-- CreateTable` comments. But the tool accepts arbitrary SQL (including piped stdin), so this could affect hand-crafted SQL. Consider adding a test documenting this known limitation.

---

### :bulb: Endpoint Regex Fragility in Test Setup

**File:** `node/prisma/test/transform.test.ts` — `connectToDsql` (~line 13)

```typescript
const match = endpoint.match(/^[a-z0-9]+\.dsql(?:-[^.]+)?\.([a-z0-9-]+)\.on\.aws$/);
```

No tolerance for trailing whitespace (common from `$(cat file)` or env var copy-paste) or uppercase. A `.trim()` on the endpoint and case-insensitive flag would make this more robust.

---

### :bulb: Duplicated `--identity-cache` Argument Parsing

**File:** `node/prisma/src/cli/index.ts`

The `--identity-cache` parsing block is duplicated in both the `migrate` and `transform` command handlers. Extract into a shared helper to keep them in sync.

---

## Positives

- :green_circle: **SERIAL regex ordering is safe** — BIGSERIAL/SMALLSERIAL replaced before SERIAL; `\b` boundaries prevent false matches on column names. Verified by tests.
- :green_circle: **identityCache range validation is correct** — Correctly enforces DSQL's `CACHE = 1 or >= 65536` constraint per [DSQL docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/sequences-identity-columns-working-with.html). Tests cover boundaries.
- :green_circle: **Warning vs error distinction is correct** — `autoincrement()` warns (DSQL supports it, hot key risk); `@db.Serial` errors (SERIAL syntax unsupported).
- :green_circle: **Clean refactoring** — `checkUnsupportedFeatures` split into `checkSerialTypes` + `checkFulltextIndex` with single responsibilities.
- :green_circle: **No ReDoS risk** — Regex patterns are simple word-boundary matches with no nested quantifiers.
- :green_circle: **OCC retry logic in integration tests** — `queryWithRetry` with exponential backoff handles DSQL's optimistic concurrency correctly.
- :green_circle: **Backwards compatible** — `autoincrement()` relaxed from error to warning; `--identity-cache` is optional with sensible default of 1.

---

## Recommended Actions

| # | Action | Priority |
|---|--------|----------|
| 1 | **Fix silent no-op tests** — use `test.skip`/`describe.skip` instead of `if (!dsql) return` across 15+ tests | Must fix |
| 2 | **Fix `--identity-cache` scientific notation bug** — `1e100` produces invalid SQL `CACHE 1e+100`; use `Number.isSafeInteger()` | Must fix |
| 3 | **Update PR description** — all SERIAL variants map to BIGINT (code is correct, description is stale) | Must fix |
| 4 | **Fix fragile `@db.Serial` substring matching** — use exact match or regex to avoid ordering dependency | Should fix |
| 5 | **Remove changelog comment** in `validate.ts` per reviewer feedback | Should fix |
| 6 | **Document behavioral changes** — `autoincrement()` error→warning, removed warnings | Should fix |
| 7 | **Log cleanup errors** instead of swallowing with `.catch(() => {})` | Should fix |
| 8 | **Add test for SERIAL in SQL comments** to document known limitation | Nice to have |
| 9 | **Trim endpoint env var** and use case-insensitive matching in test setup | Nice to have |
| 10 | **Extract `--identity-cache` parsing** into shared helper | Nice to have |

---

<sub>:robot: This review was generated by Claude Code (Opus 4.6) | Review ID: `PR-339-fdeea7910f5e9`</sub>

PR #368, opened 2026-04-02
build(deps): update prisma to 7.6.0

gxjx-x:
Update dependency prisma from 7.3.0 to 7.6.0.


anwesham-lab (approved):
why didn't this result in an upgrade on the package.json itself?

PR #401, opened 2026-04-24
feat(prisma): replace custom SQL transform with dsql-lint

amaksimo:
## Summary

- Replace ~450 lines of custom SQL parsing/transformation in `transform.ts` with [`dsql-lint`](https://github.com/awslabs/aurora-dsql-tools/tree/main/dsql-lint) from aurora-dsql-tools
- Add `dsql-lint.ts` utility for binary resolution (`DSQL_LINT_PATH` env var, then `PATH`)
- Add `lint` CLI subcommand for checking SQL without applying fixes
- Clean dsql-lint stderr output so users see diagnostic messages without temp file paths
- Keep Prisma schema validator (`validate.ts`) unchanged — dsql-lint operates on SQL, not `.prisma` files

dsql-lint covers significantly more DSQL rules than the old custom code: SERIAL types, JSON/JSONB, arrays, TEMPORARY tables, triggers, PARTITION BY, sequences, CACHE validation, multi-DDL transaction splitting, and more.

## Test plan

- [ ] 47 unit/integration tests pass (up from 41 — 6 new tests added)
- [ ] Verify `npm run dsql-migrate -- prisma/schema.prisma -o migration.sql` produces valid output
- [ ] Verify `npm run dsql-lint migration.sql` detects issues
- [ ] Verify `npm run dsql-transform input.sql -o output.sql` applies fixes
- [ ] Verify stdin piping: `cat input.sql | npm run dsql-transform`
- [ ] Verify helpful error when `dsql-lint` binary is not found
- [ ] CI: ensure `dsql-lint` binary is available (via `cargo install dsql-lint` or `DSQL_LINT_PATH`)

PR #411, opened 2026-04-27
refactor(prisma): delegate validator checks to dsql-lint

amaksimo:
## Summary

- **Replace hand-maintained Prisma schema checks with dsql-lint's lint mode.** The validator previously had custom regex checks for `autoincrement()`, `@db.Serial`, `@db.SmallSerial`, `@db.BigSerial`, `@@fulltext`, `Int @id`, `BigInt @id`, and `gen_random_uuid()` without `@db.Uuid`. These are all removed.
- **The validator now generates SQL via `prisma migrate diff` and runs `dsql-lint` on the output** to catch all SQL incompatibilities. This means the validator no longer needs manual updates when DSQL adds or removes feature support — dsql-lint is the single source of truth.
- **Only `relationMode = "prisma"` remains as a Prisma-level check** since it cannot be detected via SQL generation.
- **`dsql-migrate` skips the SQL lint step** (`skipSqlLint` option) since its transform step (`dsql-lint --fix`) already handles those issues.

Net: -269 lines, +68 lines across validator, tests, and README.

## Test plan

- [ ] `validate` command catches SERIAL (from autoincrement) via dsql-lint
- [ ] `validate` command catches CREATE INDEX without ASYNC via dsql-lint
- [ ] `validate` command reports missing `relationMode = "prisma"`
- [ ] `validate` passes for DSQL-compatible schema (no indexes, UUID IDs)
- [ ] `dsql-migrate` succeeds end-to-end (skips SQL lint, transform fixes issues)
- [ ] `dsql-migrate` fails on missing `relationMode`
- [ ] All 51 tests pass


amaksimo:
Closing — recreating on a clean branch to avoid merge conflicts from the already-merged PR #401.

PR #412, opened 2026-04-27
refactor(prisma): delegate validator checks to dsql-lint

amaksimo:
## Summary

- **Replace hand-maintained Prisma schema checks with dsql-lint's lint mode.** The validator previously had custom regex checks for `autoincrement()`, `@db.Serial`, `@db.SmallSerial`, `@db.BigSerial`, `@@fulltext`, `Int @id`, `BigInt @id`, and `gen_random_uuid()` without `@db.Uuid`. These are all removed.
- **The validator now generates SQL via `prisma migrate diff` and runs `dsql-lint` on the output** to catch all SQL incompatibilities. This means the validator no longer needs manual updates when DSQL adds or removes feature support — dsql-lint is the single source of truth.
- **Only `relationMode = "prisma"` remains as a Prisma-level check** since it cannot be detected via SQL generation.
- **`dsql-migrate` skips the SQL lint step** (`skipSqlLint` parameter) since its transform step (`dsql-lint --fix`) already handles those issues.
- **Cleaned up dead code**: removed `ValidateOptions` interface (replaced with inline boolean), removed dead `type` field from `ValidationIssue`, added error reporting for `prisma migrate diff` failures.

Net: -1132 lines removed, +612 lines added across validator, tests, and README.

## Test plan

- [x] `validate` command catches SERIAL (from autoincrement) via dsql-lint
- [x] `validate` command catches CREATE INDEX without ASYNC via dsql-lint
- [x] `validate` command reports missing `relationMode = "prisma"`
- [x] `validate` catches `@db.Serial`, `@db.SmallSerial`, `@db.BigSerial` (Prisma P1012 errors)
- [x] `validate` catches `@@fulltext` (Prisma P1012 error)
- [x] `validate` reports error when prisma cannot parse schema
- [x] `validate` passes for DSQL-compatible schema (no indexes, UUID IDs)
- [x] `dsql-migrate` succeeds end-to-end (skips SQL lint, transform fixes issues)
- [x] `dsql-migrate` fails on missing `relationMode`
- [x] All 56 tests pass

### DIFF
--- a/node/prisma/src/cli/index.ts
+++ b/node/prisma/src/cli/index.ts
@@ -21,7 +21,7 @@ Commands:
 
   validate <schema>
     Validates a Prisma schema file for DSQL compatibility.
-    Reports errors for unsupported features like autoincrement, foreign keys, etc.
+    Generates SQL and checks it with dsql-lint.
 
   transform [input] [-o output]
     Transforms SQL migrations to be DSQL-compatible using dsql-lint --fix.
@@ -177,12 +177,7 @@ Examples:
     process.exit(1);
   }
 
-  const warnings = validationResult.issues.filter((i) => i.type === "warning");
-  if (warnings.length > 0) {
-    console.log(formatValidationResult(validationResult, schemaPath));
-  } else {
-    console.log(`✓ Schema is DSQL-compatible`);
-  }
+  console.log(`✓ Schema is DSQL-compatible`);
 
   // Step 2: Generate migration using Prisma
   const fromSource = fromUrl || fromConfigDatasource ? "database" : "empty";
--- a/node/prisma/src/cli/index.ts
+++ b/node/prisma/src/cli/index.ts
@@ -71,6 +71,7 @@ async function main(): Promise<void> {
       const result = await validateSchema(schemaPath);
       console.log(formatValidationResult(result, schemaPath));
       process.exit(result.valid ? 0 : 1);
+      break;
     }
 
     case "transform": {
@@ -167,9 +168,7 @@ Examples:
 
   // Step 1: Validate schema
   console.log(`Validating ${path.basename(schemaPath)}...`);
-  const validationResult = await validateSchema(schemaPath, {
-    skipSqlLint: true,
-  });
+  const validationResult = await validateSchema(schemaPath, true);
 
   if (!validationResult.valid) {
     console.log(formatValidationResult(validationResult, schemaPath));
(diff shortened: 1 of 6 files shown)


node/prisma/src/cli/index.ts
```diff
@@ -167,20 +168,15 @@ Examples:
 
   // Step 1: Validate schema
   console.log(`Validating ${path.basename(schemaPath)}...`);
-  const validationResult = await validateSchema(schemaPath);
+  const validationResult = await validateSchema(schemaPath, true);
```
pkale, line 171: why do we need to pass in a hard coded boolean here?

node/prisma/src/cli/validate.ts
```diff
@@ -85,100 +74,50 @@ function checkRelationMode(
...
-        type: "warning",
-        message: "@id field using gen_random_uuid() should have @db.Uuid",
-        line: index + 1,
-        suggestion: "Add @db.Uuid to ensure proper UUID storage",
-      });
-    }
-  });
-}
-
-function checkUnsupportedFeatures(
```
pkale, line 136: we use to have a separate workflow for unsupported features, are we still supporting this branching with the linter? Could you provide an example of how the flow changes?

pkale (commented):
left some comments

node/prisma/src/cli/validate.ts
amaksimo: Yes, the unsupported features are still caught, they're just detected by dsql-lint now instead of hard-coded regex we had before. Steps:

1. validateSchema() generates SQL from the schema via prisma migrate diff --from-empty --to-schema <file> --script
2. It writes that SQL to a temp file and runs dsql-lint (lint mode, no --fix) on it
3. dsql-lint reports errors like Column "id" uses SERIAL, which is not supported in DSQL or CREATE INDEX without ASYNC
4.  Those errors get parsed and returned as ValidationIssues


For things that Prisma rejects at schema level before SQL generation (@db.Serial, @db.SmallSerial, @db.BigSerial, @@fulltext), the prisma migrate diff command itself fails with error: P1012 — and we now capture that error and report it too.

**The key benefit: when DSQL adds support for new features (or dsql-lint adds new rules), we don't have to update this code — dsql-lint is the single source of truth.**

node/prisma/src/cli/index.ts
amaksimo: The dsql-migrate command does both validate and transform. The true here tells validate to skip the SQL lint step, because the transform step (step 3, dsql-lint --fix) is about to handle those same SQL issues anyway. Running lint during validation would be redundant and would also fail on things like CREATE INDEX that transform knows how to fix.

The standalone validate command omits this flag, so it runs the full lint including dsql-lint.

pkale (approved):

PR #415, opened 2026-05-01
refactor(prisma): use npm-bundled dsql-lint with JSON output and stdin

amaksimo:
## Summary

Task 7 of the dsql-lint generic CLI improvement series. Consumes the features shipped in `dsql-lint 0.1.3` (now on crates.io + `@aws/dsql-lint` on npm) to replace temp-file + regex-scraping plumbing with typed JSON + stdin.

## What changed

**Packaging**
- Add `@aws/dsql-lint` as a runtime dependency. `npm install @aws/aurora-dsql-prisma-tools` now ships the `dsql-lint` binary automatically; no separate `cargo install` step.
- `findDsqlLint()` resolution order: `DSQL_LINT_PATH` override → bundled platform-specific native binary (via `require.resolve('@aws/dsql-lint-<platform>/package.json')`) → PATH scan for cargo-install backcompat.

**API consumption**
- New `runDsqlLintWithStdin(sql, args)` helper pipes SQL via stdin and parses `--format json` output into typed `DsqlLintJsonOutput` (`schema_version`, `files[]`, `summary`, per-diagnostic `fix_result.status`).
- `transform.ts` and `validate.ts` both consume the structured JSON. No more temp files. No more regex-scraping stderr. `lintMigration` now takes a SQL string directly.

**Exit code contract (matches dsql-lint --help)**
- Exit 0: clean, or all fixes applied without warnings.
- Exit 1: unfixable errors or I/O failure — surface to user, don't write output.
- Exit 3: all fixes applied, but some produced warnings (e.g. FK removal). Still a successful transform — we write the output file and propagate the code so CI can distinguish "cleanly fixed" from "review recommended."

**Text output**
- New `reportDsqlLintDiagnostics()` renders JSON diagnostics with severity labels matching dsql-lint's own text mode: `ERROR` / `WARNING` / `INFO` / `FIXED`.

## Test updates

- Exit-code expectations: FK removal → 3, `CREATE INDEX` without ASYNC → 0, `TRUNCATE` / `DROP CONSTRAINT` → 1.
- Old `expect(result.stderr).toContain("unfixable")` replaced with structured `result.output.files[].diagnostics[].fix_result.status === "unfixable"` checks.
- Obsolete "does not leak temp file paths" test removed (no more temp files).
- The `findDsqlLint() not found` test was removed — with `@aws/dsql-lint` now a direct dependency, `require.resolve` always succeeds after `npm install`, so the "not found" branch is no longer reachable from unit tests.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test -- --testPathIgnorePatterns=examples` — 58 / 58 passing (the `examples/veterinary-app` suite fails due to an unrelated pre-existing missing-prisma-generate issue that also fails on `main`).
- [ ] After merge: npm registry propagation for `@aws/dsql-lint@0.1.3` completes so `npm install` in CI pulls the bundled binary.


gxjx-x (approved):

node/prisma/README.md
```diff
@@ -26,15 +26,7 @@ npm install --save-dev @aws/aurora-dsql-prisma-tools
...
-### Prerequisites
-
-This package requires [`dsql-lint`](https://github.com/awslabs/aurora-dsql-tools/tree/main/dsql-lint) for SQL transformation and linting. Install it via:
-
-```bash
-cargo install dsql-lint
-```
-
-You can also set the `DSQL_LINT_PATH` environment variable to point to the binary location.
+`@aws/dsql-lint` is pulled in automatically as a dependency — the prebuilt native binary that ships with it is used at runtime, so there's no additional setup. If you prefer an existing installation (`cargo install dsql-lint`) or a custom path, set the `DSQL_LINT_PATH` environment variable and it will take precedence.
```
anwesham-lab: nit: verbose

rare case when I'd say AI-like language is almost better

ie:
```
`@aws/dsql-lint` is included and uses its bundled binary by default (no setup required). To use a custom or existing installation, set `DSQL_LINT_PATH`.
```

node/prisma/package.json
```diff
@@ -43,6 +43,9 @@
   "engines": {
     "node": ">=20"
   },
+  "dependencies": {
+    "@aws/dsql-lint": "^0.1.3"
```
anwesham-lab: do we want minor pins too? normally it's considered excessive no?

node/prisma/src/cli/dsql-lint.ts
```diff
@@ -23,26 +35,139 @@ function findDsqlLint(): string {
...
-export interface DsqlLintResult {
-  exitCode: number;
-  stderr: string;
+const PLATFORM_PACKAGES: Record<string, string> = {
+  "darwin-arm64": "@aws/dsql-lint-darwin-arm64",
+  "darwin-x64": "@aws/dsql-lint-darwin-x64",
+  "linux-arm64": "@aws/dsql-lint-linux-arm64",
+  "linux-x64": "@aws/dsql-lint-linux-x64",
+  "win32-x64": "@aws/dsql-lint-win32-x64",
+};
```
anwesham-lab: nit: consts should always be at the top and also I'm a fan of just having a separate constants.ts tbh

anwesham-lab:
### Code review

No ≥80-confidence blockers. Two ≥60 findings that converged across multiple reviewers in a 14-agent independent audit + 5-agent step-4 + ≥5-agent step-5 scoring (both Sonnet and Opus passes):

**1. `handleTransform` / `handleMigrate` narrow exit-code handling loses prior safety net (score 60)**

The CLI now only branches on `exitCode === 1` and `=== 3`; anything else (upstream `dsql-lint` emits 2 for clap usage errors per its `exit_code_test.rs`; Rust panic is 101; SIGKILL/OOM is 137) falls through, writes `result.sql` to disk, and exits 0. The prior code used `if (result.exitCode !== 0) process.exit(result.exitCode)` as a catch-all — this is a narrowing of that guard (PR #401, `83139de`). Suggest `exitCode !== 0 && exitCode !== 3` to preserve the original safety net.

https://github.com/awslabs/aurora-dsql-orms/blob/9885403b110537d0c2a42eff2434cb0b860d0e00/node/prisma/src/cli/index.ts#L365-L385

**2. `result.output.files[0]?.fixed_sql ?? ""` can silently yield zero-byte output (score 70)**

`DsqlLintJsonOutput` declares `files: DsqlLintFileOutput[]` and `fixed_sql: string | null` — both empty `files[]` and `null` `fixed_sql` are explicitly in-contract. Combined with finding 1, the caller can write a zero-byte migration to disk and exit 0. Guard explicitly: if `files[0]?.fixed_sql == null` on non-error exit, throw.

https://github.com/awslabs/aurora-dsql-orms/blob/9885403b110537d0c2a42eff2434cb0b860d0e00/node/prisma/src/cli/transform.ts#L540-L550

**Other observations (below ≥80 threshold, mentioned for transparency):**

- `runDsqlLintWithStdin` drops `stderr` from the return type — non-zero exits with valid JSON but advisory stderr content are now silently discarded (score 50)
- `findPlatformPackageBinary` bare `catch { return null }` hides `EACCES` / malformed `package.json` as `MODULE_NOT_FOUND` (score 50)
- `schema_version !== SUPPORTED_SCHEMA_VERSION` strict equality rejects forward-compatible additive bumps (score 50 — the author explicitly comments on this choice)
- HELP block documents exit 0/1/3 but omits 2 — doc accuracy nit (score 40)
- CHANGELOG.md not updated for the 0.1.0 → 0.1.1 bump
- Node CI (`.github/workflows/node-prisma-ci.yml`) still `cargo install dsql-lint@0.1.2` — the new npm-bundled path is not exercised by CI

False positives confirmed after scrutiny: stdout-write/exit-3 ordering (score 0), test no-throw assertion (score 0 — `expect(thrown?.status).toBe(3)` fails correctly when `thrown` is undefined).

Reviewed at HEAD `9885403b110537d0c2a42eff2434cb0b860d0e00`. Clean across git history, CLAUDE.md compliance (no CLAUDE.md exists), and code-comment compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

node/prisma/package.json
amaksimo: Fair enough

node/prisma/README.md
amaksimo: I agree

node/prisma/src/cli/dsql-lint.ts
amaksimo: Done, agreed. I think given we have just 2, no need for now. Top of file should be enough.

anwesham-lab (approved):

PR #476, opened 2026-06-08
Enable native JSONB support in adapters

amaksimo:
## Summary

DSQL now supports `JSON` and `JSONB` column types natively, so the adapter-side workarounds that compensated for their absence are no longer needed.

- **Django**: drop `has_native_json_field = False` so Django's `JSONField` works without the override.
- **SQLAlchemy**: restore the upstream `_columns_query` body so `identity_options` reflection uses `json_build_object` instead of returning NULL. Verified `json_build_object`, `pg_sequence`, and `pg_get_serial_sequence` all work in DSQL today.
- **SQLAlchemy README**: drop the "datatype json not supported" Dialect Features bullet.
- **Tortoise**: drop the `JSONField not supported` section from `KNOWN_ISSUES.md`.
- **Tortoise**: drop the Aerich `content` JSON→TEXT patch. Stock aerich uses `fields.JSONField(encoder, decoder)`, which now creates a JSONB column on DSQL. Updated the test, the README, and `ADAPTER_BEHAVIOR.md` to match.

### Backward compatibility — Aerich migration table

Existing users have an `aerich` table whose `content` column is TEXT (created under the prior patch). After this change, `aerich init-db` against a fresh DSQL cluster creates `content` as JSONB. Existing TEXT-typed tables continue to work at runtime: Tortoise's `JSONField` reads/writes JSON-encoded text unchanged regardless of column type. Users who want to convert can do so manually; we don't ship an automatic migration.

## Test plan

- [x] Django unit tests (`test_features.py`) pass after dropping the `has_native_json_field` override and the corresponding assertions.
- [x] Tortoise unit tests pass (119 tests, including the simplified `test_aerich_compat.py`).
- [x] SQLAlchemy `_columns_query` compiles and emits `json_build_object` (verified via local smoke test).
- [ ] SQLAlchemy integration tests against a live DSQL cluster (please run via CI).
- [ ] Tortoise integration tests against a live DSQL cluster (please run via CI).
---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/aurora-dsql-orms/blob/main/LICENSE).

### DIFF
--- a/python/django/aurora_dsql_django/features.py
+++ b/python/django/aurora_dsql_django/features.py
@@ -37,9 +37,6 @@ class DatabaseFeatures(features.DatabaseFeatures):
     # Does the database support deferrable unique constraints?
     supports_deferrable_unique_constraints = False
 
-    # Does the database have native JSON field support?
-    has_native_json_field = False
-
     # Can the database introspect materialized views?
     can_introspect_materialized_views = False
 
--- a/python/django/aurora_dsql_django/tests/unit/test_features.py
+++ b/python/django/aurora_dsql_django/tests/unit/test_features.py
@@ -36,9 +36,6 @@ class TestDatabaseFeatures(unittest.TestCase):
     def test_supports_deferrable_unique_constraints(self):
         self.assertFalse(self.features.supports_deferrable_unique_constraints)
 
-    def test_has_native_json_field(self):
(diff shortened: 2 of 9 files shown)


praba2210 (approved):
Looks good!

PR #479, opened 2026-06-14
feat(hibernate): Hibernate 7.x compatibility (v2.0.0)

rmconstantin:
## Summary

Updates `aurora-dsql-hibernate-dialect` to 2.0.0 for Hibernate 7.2+ / Spring Boot 4.x. v1.0.x fails at runtime on Hibernate 7.x (`NoClassDefFoundError: org/hibernate/dialect/PgJdbcHelper`) due to ORM 7.0 class relocations.

## Changes

**Dialect**
- Import relocations: `PgJdbcHelper`, `PostgreSQLCasting*`, `PostgreSQLUUIDJdbcType` → `dialect.type`; `PostgreSQLSqlAstTranslator` → `dialect.sql.ast`; `TemporalUnit`, `FetchClauseType` → `query.common`
- `JavaTypeRegistry.getDescriptor()` → `resolveDescriptor()`
- Inline `PG_UUID_JDBC_TYPE` → `PostgreSQLUUIDJdbcType.INSTANCE`
- Removed custom `LockingStrategy` override (`Lockable` removed in Hibernate 7; DSQL is OCC-only)
- build: `hibernate-core:7.2.12.Final`, Java 17

**Sample app (pet-clinic)**
- Migrated to Spring Boot 4.0.7: starter/package relocations (`webmvc`, `restclient`), `@MockBean` → `@MockitoBean`, property renames
- dialect 2.0.0 loaded via locally-built JAR (not yet on Maven Central)

## Compatibility

| Dialect | Hibernate ORM | Spring Boot |
|---|---|---|
| 1.0.x | 6.6.x | 3.x |
| 2.0.0 | 7.2+ | 4.x |

## Testing
- 38 unit tests pass; clean `gradle clean build`
- Integration tests run with `RUN_INTEGRATION=TRUE` (live cluster)
- 5 deprecation warnings, intentional (methods still functional in 7.2+)

## Breaking change

Major version bump. Stay on v1.0.x for Spring Boot 3.x / Hibernate 6.6.x.

PR #499, opened 2026-06-23
ci: weekly compatibility canary for all ORM adapters

amaksimo:
## Summary

Proactively detect upstream framework breaking changes — like the Hibernate 7 `PgJdbcHelper` relocation that broke the dialect at a customer's runtime (#479) — *before* a customer hits them. Applies uniformly to all five ORM adapters (Hibernate, Django, SQLAlchemy, Tortoise, Prisma).

## Why nothing warned us last time

1. **CI only ever tested the pinned version.** The dialect compiles `compileOnly` against a frozen Hibernate; the Python/Node adapters test against locked versions. Green CI meant "works with what we froze," never "works with what customers will pull."
2. **Dependabot was muting the signal.** `dependabot.yml` ignored `hibernate-core >=7` / `spring-boot >=4`, so the one mechanism that would have gone red was silenced (reasonably — to keep the stable build green).

The `PgJdbcHelper` failure is a **linkage/compile-time** break (class relocation → `NoClassDefFoundError` at `EntityManagerFactory` build). That class of break is the cheapest thing to detect: recompiling against the new major surfaces it with **no live cluster needed**.

## What this adds

- **`.github/workflows/compat-canary.yml`** — weekly (Mon 06:00 UTC) + manual. One independent job per ORM that builds/tests the adapter against the **latest** framework release (incl. pre-releases) instead of the pinned version. `continue-on-error` semantics via per-job reporting — it **never gates PRs or releases**; it is not part of ci-gate. Tier-1 scope only: recompile + unit/import probes, **no live DSQL cluster** (the break class is linkage, not SQL semantics).
  - SQLAlchemy's suite requires a cluster, so it gets an import-only smoke probe that still exercises the relocation-prone PG dialect imports.
- **`.github/actions/report-canary-failure/`** — composite action that opens (or comments on) a single dedup'd `compat-canary` tracking issue per ORM on failure.
- **`build.gradle`** — `hibernateVersion` is now overridable via `-PhibernateVersion=…` (default unchanged at `7.4.2.Final`). Verified: override flows through, default falls back.
- **`dependabot.yml`** — the `>=7` / `>=4` ignores were **stale** (the repo already moved to Hibernate 7.4.2 / Spring Boot 4, and they referenced the old `org.hibernate:hibernate-core` coordinate, now `org.hibernate.orm`). Rolled forward to next-major (`>=8` / `>=5`) and fixed the coordinate so in-major updates flow again and the canary owns the next-major signal. Dropped the removed `hibernate-ant` entry.

## How it behaves

- Framework ships a breaking major → next weekly canary goes red → a tracking issue is filed automatically. Stable releases keep building against pinned versions, unaffected.
- No new live-cluster cost; jobs are independent so one ORM breaking doesn't mask others.

## Testing

- All three YAML files parse (`yaml.safe_load`).
- Gradle override verified locally: `-PhibernateVersion=9.9.9.PROBE` resolves through; no-arg falls back to `7.4.2.Final`.
- Cluster-coupling audited per ORM to confirm each probe runs cluster-free (Django/Tortoise unit tests + Prisma jest don't connect; SQLAlchemy → import probe).

## Follow-ups (deliberately not in this PR)

- Add tier-2 (in-memory/H2 sample-app boot vs latest) if a future break slips past compile/unit — catches Spring-Boot starter relocations.
- Add live-DSQL integration to the canary only if a framework introduces a SQL-semantic break that tier-1 misses.


amaksimo:
## Self-review (Iteration 1)

Dispatched 5 parallel general-purpose review sub-agents covering the procedure's required slots (shallow-bug + GHA safety; drift / PR-body / commit audit; security; code-quality + simplification + comment-compliance; test-realism + idiom + repo-conventions). Slots 9 and 18 had no work to do — no prior review comments. After de-duplication and head-SHA verification (originally `8e20dbe`, advanced to `a620cd1` mid-review) the table below captures every finding ≥ 60 confidence.

| # | Confidence | Area | Finding | Disposition | Commit / SHA |
|---|---|---|---|---|---|
| 1 | 70 | GHA / dedup logic | `gh issue list --search "in:title \"[compat-canary] …\""` — GitHub search tokenizes/strips punctuation, so the bracketed prefix may not match and a new issue gets filed every Monday despite an open one already existing. The `--jq 'select(.title == …)'` post-filter only operates over what search returns. | **fixed** — drop `--search`, dedup via `--label compat-canary` + `jq env.TITLE` exact-match. | f0eda1a |
| 2 | 90 | comment accuracy | `build.gradle:24` comment said "the compat-canary workflow probes the next major" but `-PhibernateVersion=+` resolves to the latest available version — could be a 7.x point release if no 8.x exists. | **fixed** — change "next major" → "latest available version". | f0eda1a |
| 3 | 85 | comment | `--no-sync` after `uv pip install --upgrade` is intentional (without it, `uv run` re-syncs from uv.lock and reverts the upgrade). One comment near the first occurrence (django job) covers all three Python jobs. | **fixed** — added explanatory comment. | f0eda1a |
| 4 | 65 | jq quoting | `--jq "select(.title == \"${TITLE}\")"` — a future ORM containing `"` or `\` in its name would break the jq expression. | **fixed** (folded into #1) — `env.TITLE` is safe against any string. | f0eda1a |
| 5 | 88 | YAML / DRY | `permissions: issues: write` repeated on all 5 jobs while `permissions: {}` is set workflow-level. | **false-positive: prefer per-job least-privilege** — top-level `{}` + per-job grant is the GitHub-recommended baseline; hoisting widens scope to any future helper job that doesn't need it. Verbosity is acceptable. | n/a |
| 6 | 92 | rendering | Two sub-agents flagged BODY indented inside the YAML block scalar would render as a Markdown code block. | **false-positive: verified** — confirmed via `yaml.safe_load` that the `\|` block scalar strips the 8-space common indent before bash sees the script. BODY content is left-aligned at runtime. | n/a |
| 7 | 80 | YAML / DRY | 3 Python jobs collapse into a `strategy.matrix.include` (~50-line saving). | **deferred: low-value** — probe commands and env vars differ enough per ORM that the matrix shape obscures more than it saves. Each job stays grep-friendly when something fails. Re-evaluate if a 6th Python ORM is added. | n/a |
| 8 | 75 | accuracy nit | Prisma `npm install prisma@latest` excludes pre-releases by npm convention, while the PR body says canaries probe pre-releases. The four other jobs do; Prisma alone doesn't. | **deferred** — npm has no general "latest including pre-releases" tag; switching to `@next` excludes regular releases. The current command catches Prisma majors at GA, sufficient for tier-1. | n/a |
| 9 | 70 | dependabot | Spring Boot ignore moved from `>=4` to `>=5` — re-opens dependabot for Spring Boot 4.x updates. | **false-positive: intentional** — PR description explicitly says "Rolled forward to next-major (`>=8` / `>=5`) … so in-major updates flow again". | n/a |
| 10 | 78 | comment hedge | Dependabot "Next-major Hibernate **likely** has breaking API changes" — replace with "may introduce" or remove. | **deferred: cosmetic** — current wording is acceptable. | n/a |
| 11 | 60 | shell | `set -euo pipefail` + `head -n1` could trip SIGPIPE-141. | **false-positive: verified** — `gh issue list --json` produces small output that head consumes fully; `head -n1` exits 0 cleanly; pipefail doesn't fire. | n/a |

**Net:** 4 of 11 findings landed as one fixup commit (3 distinct fixes, #4 folded into #1). The rest are either verified false positives or deferred as cosmetic / intentional.

Iteration 2 (mandatory validation pass over the new HEAD `f0eda1a`) is in progress; will follow up if it surfaces anything new.

amaksimo:
## Self-review (Iteration 2 — validation pass)

Fresh review pass on HEAD `f0eda1a`. Adversarial scrutiny of each Iteration 1 fix plus a sweep for adjacent issues the original findings might have masked.

**Verdict:** 0 new high-confidence (≥ 60) findings. No regressions. Convergence reached.

| Risk vector probed | Result |
|---|---|
| Dedup rewrite — does `gh issue list --label compat-canary` error on the FIRST-ever run when the label doesn't exist yet? | **Safe.** Returns `[]` exit 0; pipefail not tripped. Verified against this repo (label currently does not exist). |
| Dedup rewrite — does `env.TITLE` resolve in `gh`'s embedded gojq? | **Safe.** gojq supports `env.X` like jq 1.6+. |
| Worst case: `gh label create … 2>/dev/null \|\| true` swallows a transient failure, then `gh issue create --label compat-canary` errors because the label is missing. | **Visible failure**, not silent dedup-bypass. The composite-action step fails loudly in the workflow run; the existing `if: failure() && inputs.dry-run != true` path on the **caller** would still execute (it's gated only on the probe step's failure, not the report step). Acceptable for a defensive canary. Both `gh label create` and `gh issue create` use the same `issues: write` scope, so the most likely cause of label-create failure is the rare API blip rather than a permission problem. |
| `inputs.dry-run != true` on the cron trigger (the `inputs` context is empty for `schedule`). | **Safe.** Per GitHub Actions coercion (null→0, true→1), `null != true` → `true`. Scheduled runs file issues as designed. |
| Comment-only changes (build.gradle wording, compat-canary.yml `--no-sync` annotation). | **Pure prose**, no executable behavior change. |
| Adjacent: any past canary issue filed without the `compat-canary` label would be invisible to the new label-scoped dedup. | **Not applicable** — this is the very first deployment of the action; no historical issues exist to worry about. |

**Convergence (Section 7):** 2 iterations completed; Iteration 2 reports 0 new high-confidence issues; no previously fixed item regressed. Self-review is complete.

anwesham-lab (approved):

anwesham-lab:
LGTM!

PR #500, opened 2026-06-24
feat(ef-core): add EF Core 9.x adapter for Aurora DSQL

praba2210:
## Summary
- Adds EF Core 9.x adapter for Aurora DSQL with connection wrapping, IAM auth via DsqlDataSource, OCC retry via DsqlExecutionStrategy, and UUID/IDENTITY primary key conventions
- Native migration support via DsqlMigrator with dsql-lint integration for DDL transformation (CREATE INDEX ASYNC, IF NOT EXISTS idempotency)
- InventoryApi sample app demonstrating CRUD, OCC retry, batch operations, navigation properties, and migrations
- CI integration (per-adapter GitHub Actions ci/release workflows + ci-gate wiring), Apache-2.0 license headers on all .cs files, and a .NET pre-commit license hook
- Unit tests (23) and integration tests (54) covering CRUD, queries, transactions, execution strategy, connections, type mapping, raw SQL, constraints, and migrations

## Test plan
- [x] Unit tests pass (`dotnet test .../UnitTests/`)
- [x] Integration tests pass against live DSQL cluster (`CLUSTER_ENDPOINT=... dotnet test .../IntegrationTests/`)
- [x] InventoryApi example integration tests pass (`dotnet test .../InventoryApi.IntegrationTests/`)


dotnet/ef-core/README.md
```diff
@@ -0,0 +1,314 @@
...
+  Other levels are logged and handled gracefully.
+
+- **Connection Pooling** — Npgsql pooling via DbDataSource.
+
+- **IAM Authentication** — Via `Amazon.AuroraDsql.Npgsql`.
+
+- **Batch Operations** — Multi-entity `SaveChanges` works
+  seamlessly; savepoints suppressed transparently.
+
+### Design Considerations
```
amaksimo: Remove this section

dotnet/ef-core/README.md
```diff
@@ -0,0 +1,314 @@
...
+- **Referential Integrity** — DSQL delegates FK enforcement to
+  the application layer. Navigation properties, `Include`,
+  `ThenInclude`, and LINQ joins work normally — enforce
+  consistency in your service layer.
+
+- **Primary Keys** — See [Primary Keys](#primary-keys) above.
+
+- **Migrations** — See [Migrations](#migrations) above.
+
+## Testing
```
amaksimo: Remove this section

dotnet/ef-core/README.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+Named profile:
+```csharp
+services.AddDsqlDataSource(
+    "your-cluster.dsql.us-east-1.on.aws",
+    config => config.Profile = "AwsProfile");
+```
+
+### Debug Logging
```
amaksimo: Refactor this section as showing how to enable logging. I also think the example log is super odd... looks like something from dsql-lint, not EF Core.

dotnet/ef-core/README.md
```diff
@@ -0,0 +1,314 @@
...
+## Resources
+
+- [Aurora DSQL docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/what-is-aurora-dsql.html)
+- [Concurrency control](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-concurrency-control.html)
+- [Unsupported PG features](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-unsupported-features.html)
+- [Sequences & identity columns](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/sequences-identity-columns-working-with.html)
+- [Aurora DSQL Tools](https://github.com/awslabs/aurora-dsql-tools)
+- [dsql-lint](https://github.com/awslabs/aurora-dsql-tools/tree/main/dsql-lint)
+- [EF Core docs](https://learn.microsoft.com/ef/core/)
+- [Amazon.AuroraDsql.Npgsql](https://github.com/awslabs/aurora-dsql-npgsql)
```
amaksimo: this is wrong link. Please double check each link to make sure it does not 404.

dotnet/ef-core/README.md
```diff
@@ -0,0 +1,314 @@
...
+```csharp
+builder.Services.AddDbContext<MyDbContext>(
+    (sp, options) => options.UseDsql(sp, dsql =>
+        dsql.EnableIdentityColumns()));
+```
+
+With a custom cache size (default is 65,536):
+```csharp
+options.UseDsql(sp, dsql =>
+    dsql.EnableIdentityColumns(cacheSize: 1000));
```
amaksimo: this example will not work as cache needs to be between 1 and 65535 (or something similar to that)

dotnet/ef-core/README.md
```diff
@@ -0,0 +1,314 @@
...
+```bash
+dotnet ef migrations add CreateOrders
+dotnet ef database update
+```
+
+The adapter uses [dsql-lint](https://github.com/awslabs/aurora-dsql-tools/tree/main/dsql-lint) to transform
+EF Core-generated DDL into DSQL-compatible SQL (e.g. `CREATE INDEX` →
+`CREATE INDEX ASYNC`).
+
+Because DSQL allows at most one DDL statement per transaction, migrations are
```
amaksimo: I think we should refactor this.

We should have a section near the top that discusses limitations.

Here we can mention things like not atomic, etc. Also we can use more positive framing instead of "because dsql allows at most one DDL..."

dotnet/ef-core/README.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+
+See [EF Core configuration docs](https://learn.microsoft.com/ef/core/dbcontext-configuration/)
+for more options.
+
+### Database Connection
+
+The adapter uses
+[Amazon.AuroraDsql.Npgsql](https://github.com/awslabs/aurora-dsql-npgsql)
```
amaksimo: link does not exist..

amaksimo (commented):
This PR is missing edits to the CI workflow.

We should be testing all the example code we add.

We should also be testing the adapter in CI against a real dsql-cluster. We need a proper end to end integration test that mocks a real customer workflow.

Ideally we should test a workflow which would have dsql-lint modify/transform the schema that ef core generates as well.

dotnet/ef-core/README.md
amaksimo: confirmed — `awslabs/aurora-dsql-npgsql` 404s. the package actually lives at https://github.com/awslabs/aurora-dsql-connectors/tree/main/dotnet/npgsql. same dead link is repeated at L310 in Resources, so both need updating.

dotnet/ef-core/README.md
amaksimo: same broken `aurora-dsql-npgsql` link as L73 — correct URL is https://github.com/awslabs/aurora-dsql-connectors/tree/main/dotnet/npgsql. checked the rest: every other link in the README resolves, just these two.

dotnet/ef-core/README.md
amaksimo: close — checked the docs, DSQL only accepts `CACHE = 1` or `CACHE >= 65536`, so `1000` is invalid (and the table DDL at L185-189 has the same `(CACHE 1000)`). the `65,536` default mentioned above is fine. we should also make `EnableIdentityColumns` reject `2..65535` — right now [DsqlDbContextOptionsBuilder.cs:48-52](https://github.com/awslabs/aurora-dsql-orms/blob/623a6272a4b87f4f0ab91ba9380e3efd761797ae/dotnet/ef-core/src/Amazon.AuroraDsql.EntityFrameworkCore/Infrastructure/DsqlDbContextOptionsBuilder.cs#L48-L52) only rejects `< 1`, so the bad value flows straight through to the generated `CACHE` clause.

dotnet/ef-core/src/Amazon.AuroraDsql.EntityFrameworkCore/Infrastructure/DsqlSqlTransform.cs
```diff
@@ -0,0 +1,150 @@
...
+    [GeneratedRegex(@"\bCREATE\s+TABLE\s+(?!IF\s+NOT\s+EXISTS\b)",
+        RegexOptions.IgnoreCase)]
+    private static partial Regex CreateTableRegex();
+
+    [GeneratedRegex(@"\bCREATE\s+(UNIQUE\s+)?INDEX\s+(ASYNC\s+)?(?!IF\s+NOT\s+EXISTS\b)",
+        RegexOptions.IgnoreCase)]
+    private static partial Regex CreateIndexRegex();
+
+    /// <summary>
+    /// Rewrites CREATE TABLE and CREATE [UNIQUE] [ASYNC] INDEX statements to include
```
amaksimo: doc says `CREATE [UNIQUE] [ASYNC] INDEX` but the DSQL grammar is `CREATE [UNIQUE] INDEX ASYNC` — `ASYNC` goes after `INDEX`, not before. the regex just below is correct, it's only the comment that's off.

dotnet/ef-core/src/Amazon.AuroraDsql.EntityFrameworkCore/Infrastructure/DsqlExecutionStrategy.cs
```diff
@@ -0,0 +1,90 @@
...
+            return false;
+
+        var current = exception;
+        while (current is not null)
+        {
+            if (OccRetry.IsOccError(current))
+            {
+                _logger?.LogDebug(
+                    DsqlEventId.OccRetry,
+                    "Retrying operation due to OCC conflict (SqlState 40001). Attempt {Attempt}.",
```
amaksimo: the message only names `SqlState 40001`, but `IsOccError` also matches OC000/OC001 (per the class doc on L7). reads as if only 40001 gets retried — minor, but worth widening the text so the logs aren't misleading during an OCC storm.

amaksimo:
## Review summary

Did a deep pass over the adapter. Overall it's in good shape — the wrapper/transform/retry design is sound and the test coverage is broad. Below is what I'd want addressed before merge, split into things flagged inline and a few larger items worth a decision.

### Flagged inline (objective fixes)

| # | Where | Issue |
|---|---|---|
| 1 | README L73 + L310 | `aurora-dsql-npgsql` link 404s — package is at [`aurora-dsql-connectors/dotnet/npgsql`](https://github.com/awslabs/aurora-dsql-connectors/tree/main/dotnet/npgsql). Every other README link checks out. |
| 2 | README L171, L185-189 + `DsqlDbContextOptionsBuilder.cs:48` | `cacheSize: 1000` / `(CACHE 1000)` is invalid — DSQL only allows `CACHE = 1` or `>= 65536`. The `65,536` default is fine; the builder should reject `2..65535` instead of only `< 1`. |
| 3 | `DsqlSqlTransform.cs:131` | XML-doc has `[ASYNC] INDEX`; grammar is `INDEX ASYNC`. Regex is correct, comment isn't. |
| 4 | `DsqlExecutionStrategy.cs:70` | OCC retry log names only `SqlState 40001`, but detection also covers OC000/OC001. |

Your 7 earlier README comments (remove Design-Considerations §L209 / Testing §L223, refactor Debug-Logging §L270, both dead links, the cache value, and the Limitations-section + positive-framing ask at §L89) are all still open at this HEAD — none landed in a later commit.

### Larger items — worth a decision, not blocking inline

- **No CI wiring.** The whole `dotnet/ef-core/` subtree isn't in `ci-gate.yml` and there's no `dotnet-ef-core-ci.yml` / `-release.yml`. Every sibling adapter has both, so right now these 70 tests never run on PRs and there's no release path. This is the one I'd most want sorted before merge.
- **License headers.** None of the 51 new `.cs` files carry the Apache-2.0 / SPDX header that the Amazon-authored Java/Python sources do. Worth adding a `dotnet`-scoped insert-license pre-commit hook too.
- **`OnFirstExecution()` override** ([`DsqlExecutionStrategy.cs:50`](https://github.com/awslabs/aurora-dsql-orms/blob/623a6272a4b87f4f0ab91ba9380e3efd761797ae/dotnet/ef-core/src/Amazon.AuroraDsql.EntityFrameworkCore/Infrastructure/DsqlExecutionStrategy.cs#L50-L55)) is empty, so it drops the base's `ExceptionsEncountered.Clear()`. Fine for the per-operation strategy the DI path creates, but a *reused* strategy instance (`CreateExecutionStrategy()` then two `ExecuteAsync` calls) would carry conflict counts across operations and mis-fire `MaxRetryCount`. Worth either calling `base` or clearing explicitly + a reuse test.
- **Sync-over-async default.** `AddDsqlDataSource(host)` / `UseDsql(host)` block on `CreateAsync().GetAwaiter().GetResult()`, and that's the path the quick-start and sample use. The pre-created-`DsqlDataSource` overloads are the safe ones — might be worth steering the docs/sample toward those.

---

<sub>Self-review: 2 iterations, 19 review lanes + adversarial/refute verification. Functional claims (cache range, `CREATE INDEX ASYNC` grammar, OC000/OC001, the 404) checked against AWS DSQL docs, the dsql-lint source, and live `gh`/`curl`; line numbers verified at HEAD `623a627`. A handful of agent findings were dropped as false positives (a redundant-`AsQueryable()` call that would actually break the build; "silent failures" in `DsqlLintRunner` that are unreachable against the real lint binary; and a lint exit-code comment that's actually correct). ~64 low-severity style/simplification nits were left out as author's-discretion.</sub>

praba2210:
## Review fixes — addressed at `152ae1b`

 All inline + larger items are addressed, plus a fresh self-review on top. Summary below.

### Your flagged items

| # | Where | Status |
|---|---|---|
| 1 | README L73/L310 dead `aurora-dsql-npgsql` link | ✅ Fixed — now points to [`aurora-dsql-connectors/dotnet/npgsql`](https://github.com/awslabs/aurora-dsql-connectors/tree/main/dotnet/npgsql). Re-checked every README link resolves. |
| 2 | `cacheSize: 1000` invalid + builder only rejected `< 1` | ✅ Fixed — README + sample DDL use valid values; `EnableIdentityColumns` now rejects `2..65535` (accepts `1` or `>= 65536`), with 7 added unit tests. |
| 3 | `DsqlSqlTransform.cs` doc `[ASYNC] INDEX` | ✅ Fixed — doc now reads `CREATE [UNIQUE] INDEX [ASYNC]`. |
| 4 | OCC retry log named only `40001` | ✅ Fixed — message now names `40001/OC000/OC001`. |
| 5 | README: drop Design-Considerations + Testing, refactor Logging, add Limitations w/ positive framing | ✅ Done — "Features and Limitations" near the top; Logging section shows how to enable EF Core debug logging. |
| — | No CI wiring | ✅ Added `dotnet-ef-core-ci.yml` (build + unit + examples + live-cluster integration, incl. a dsql-lint-transform end-to-end path) and `dotnet-ef-core-release.yml`, wired into `ci-gate.yml`. |
| — | License headers | ✅ Apache-2.0 / SPDX header on all `.cs` files + a `dotnet`-scoped `insert-license` pre-commit hook. |
| — | `OnFirstExecution()` empty override | ✅ Now clears `ExceptionsEncountered` explicitly so a reused strategy can't carry conflict counts across operations. |
| — | Sync-over-async default | ✅ Doc-steered: `UseDsql(host)` / `AddDsqlDataSource(host)` XML docs now point to the pre-created `DsqlDataSource` overload for deadlock-safe registration. (No deadlock in ASP.NET Core — no `SynchronizationContext`, and the blocking call runs lazily in a thread-pool DI factory — so the host overloads remain supported for the quick-start.) |

### Self-review pass (additional fixes)

Ran a multi-lane self-review over the updated branch. Confirmed fixes worth landing:

| # | Conf | Area | Finding | Disposition |
|---|---|---|---|---|
| 7 | 60 | Docs | `UseDsql(host)` pointed at `AddDsqlDataSource()`, itself sync-over-async; the genuinely safe path is the pre-created `DsqlDataSource` overload | fixed — doc now names the correct overload |
| 1 | 90 | PR body | "16 unit tests" stale → **23**; body omitted CI/license/pre-commit work | fixed — body updated |
| 3 | 75 | Dead code | `IDsqlSqlTransform.Transform` / `TransformBatch` had zero callers | fixed — removed (only `TransformBatchWithDiagnostics` is used) |
| 10 | 85 | Cleanup | `using`s duplicating `GlobalUsings.cs` | fixed — removed redundant directives |

Items reviewed and intentionally **not** changed (false positives / out-of-scope, happy to discuss):
- **Seal `DsqlExecutionStrategy`** — left unsealed deliberately; it's an EF Core extension point.
- **`DsqlLintRunner` stdin ordering / OCC jitter clamp** — verified not reachable / valid backoff design.
- **`OnFirstExecution` reuse test** — DI creates a fresh strategy per operation, so the reuse path is near-theoretical; skipped to keep scope minimal.
- **CHANGELOG / THIRD-PARTY** — CHANGELOG auto-generates at release (already wired in `dotnet-ef-core-release.yml`); THIRD-PARTY is SBOM-tooling output, not hand-authored.

**Verification:** **89/89** tests pass (23 unit + 54 integration + 12 example) against a live DSQL cluster.

dotnet/ef-core/README.md
```diff
@@ -0,0 +1,242 @@
...
+  (`SET TRANSACTION ISOLATION LEVEL`, `SAVEPOINT`, `LOCK TABLE`) at the
+  ADO.NET layer, logged at Debug level. Batch `SaveChanges` works seamlessly.
+- **Concurrency Control** — DSQL uses a single optimistic concurrency control
+  isolation level; requested isolation levels are ignored. Conflicting
+  transactions are retried automatically. See [Concurrency Control](#concurrency-control).
+- **Migrations** — DSQL applies one DDL statement per transaction, so a
+  multi-statement migration is not applied atomically. The adapter makes DDL
+  idempotent (`CREATE TABLE IF NOT EXISTS`) so a migration can be safely re-run
+  after fixing a failure. See [Migrations](#migrations).
+- **Referential Integrity** — DSQL does not enforce foreign keys; navigation
```
amaksimo, line 25: We will need to update this once we add foreign key support. 

Can you update the runbook to add this to the tasks?

amaksimo (approved):
lgtm, thanks, please make sure to release it, test locally & close out the appsec ticket + update runbook!

PR #528, opened 2026-07-22
feat(orms): support CHECK constraints (NOT VALID + VALIDATE ASYNC) in Django, SQLAlchemy, Hibernate

agarwal-govind:
## Summary

DSQL now supports adding `CHECK` constraints to existing tables via `ALTER TABLE ... ADD CONSTRAINT ... NOT VALID` followed by an asynchronous `ALTER TABLE ASYNC ... VALIDATE CONSTRAINT`, and supports inline `CHECK` constraints at `CREATE TABLE`. This PR enables real `CHECK` constraint support across the Django, SQLAlchemy, and Hibernate adapters, which previously skipped or disabled it.

### Django (`python/django`)
- `features.py`: `supports_table_check_constraints = True`.
- `schema.py`: `sql_create_check` emits `ADD CONSTRAINT ... CHECK (...) NOT VALID`; `add_constraint` issues `ALTER TABLE ASYNC ... VALIDATE CONSTRAINT` as a follow-up so existing rows are validated. Removed the old skip logic (inline CREATE TABLE checks and `DROP CONSTRAINT` are supported).
- `reference/ADAPTER_BEHAVIOR.md`: replaced the "check constraints skipped" section with async-validation behavior and the `NOT VALID` caveat.

### SQLAlchemy (`python/sqlalchemy`)
- `base.py`: override `visit_add_constraint` to append `NOT VALID` for `CheckConstraint` (non-CHECK unaffected). `supports_alter` stays `False` (it also gates FK/PK-via-ALTER auto-ordering, still unsupported). The async `VALIDATE CONSTRAINT` step is run separately (e.g. `op.execute(...)` in Alembic), documented in the README.
- Inline `CHECK` at `CREATE TABLE` already worked and is unchanged.

### Hibernate (`java/hibernate`)
- `AuroraDSQLDialect.java`: `supportsColumnCheck()` returns `true` (inline column CHECK at CREATE TABLE).

## Behavior caveat

`VALIDATE CONSTRAINT ASYNC` returns immediately; validation of existing rows runs as an async DDL job. The constraint is enforced on new writes right away. If existing rows violate it, the async job fails and the constraint stays `NOT VALID` (track via `sys.jobs`).

## Testing

- Unit tests pass: Django (`pytest` schema + features), SQLAlchemy compiler assertions, Hibernate (`gradlew test`).
- **Verified end-to-end against a gamma DSQL cluster** for all three: inline CREATE TABLE checks; post-creation `ADD CONSTRAINT ... NOT VALID` + `VALIDATE CONSTRAINT ASYNC` (job `submitted` → `completed`, constraint enforced); async failure on dirty data (`sys.jobs` status `failed`, constraint stays `NOT VALID`); and confirmed a plain `ADD CONSTRAINT ... CHECK` without `NOT VALID` is rejected by DSQL.

---

**Part of the `VALIDATE CONSTRAINT ASYNC` rollout across the DSQL ecosystem:**
- dsql-lint: https://github.com/awslabs/aurora-dsql-tools/pull/112
- agent-plugins (canonical steering): https://github.com/awslabs/agent-plugins/pull/238
- mcp (mirror steering): https://github.com/awslabs/mcp/pull/4300
- aurora-dsql-orms (this PR)

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/aurora-dsql-orms/blob/main/LICENSE).

### DIFF
--- a/java/hibernate/dialect/src/main/java/software/amazon/dsql/hibernate/dialect/AuroraDSQLDialect.java
+++ b/java/hibernate/dialect/src/main/java/software/amazon/dsql/hibernate/dialect/AuroraDSQLDialect.java
@@ -311,7 +311,8 @@ public class AuroraDSQLDialect extends Dialect {
 
   @Override
   public boolean supportsColumnCheck() {
-    return false;
+    // DSQL supports inline column CHECK constraints at CREATE TABLE.
+    return true;
   }
 
   // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/java/hibernate/dialect/src/main/java/software/amazon/dsql/hibernate/dialect/AuroraDSQLDialect.java
+++ b/java/hibernate/dialect/src/main/java/software/amazon/dsql/hibernate/dialect/AuroraDSQLDialect.java
@@ -311,7 +311,6 @@ public class AuroraDSQLDialect extends Dialect {
 
   @Override
   public boolean supportsColumnCheck() {
-    // DSQL supports inline column CHECK constraints at CREATE TABLE.
     return true;
   }
 
--- a/java/hibernate/dialect/src/test/java/software/amazon/dsql/hibernate/dialect/AuroraDSQLDialectTest.java
+++ b/java/hibernate/dialect/src/test/java/software/amazon/dsql/hibernate/dialect/AuroraDSQLDialectTest.java
@@ -87,7 +87,7 @@ public class AuroraDSQLDialectTest {
 
   @Test
   public void testColumnCheck() {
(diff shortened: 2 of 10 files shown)


java/hibernate/dialect/src/main/java/software/amazon/dsql/hibernate/dialect/AuroraDSQLDialect.java
```diff
@@ -311,7 +311,8 @@ public int getMaxVarcharLength() {
 
   @Override
   public boolean supportsColumnCheck() {
-    return false;
+    // DSQL supports inline column CHECK constraints at CREATE TABLE.
```
amaksimo: We do not need this comment.

PR #563, opened 2026-08-12
Add Aurora DSQL adapter for Drizzle ORM

praba2210:
**Add `@aws/aurora-dsql-drizzle` — a Drizzle ORM adapter for Amazon Aurora DSQL.**

It rides on `drizzle-orm/node-postgres` over the Aurora DSQL connector (a
`pg.Pool` with IAM token auth), so there's no custom dialect to maintain — just
the thin pieces DSQL actually needs: a `drizzle()` factory, an opt-in OCC-retry
transaction helper, a resumable migrator that applies one DDL per transaction,
and an `aurora-dsql-drizzle` CLI that rewrites migrations for DSQL via
dsql-lint. Ships with a runnable veterinary-app example and CI/release
workflows. Verified with 133 mocked unit tests and 18 live integration tests
against a real cluster.

The npm badges 404 until the first release publishes `0.1.0`.

## What's included

- `drizzle()` factory building an IAM-authenticated `AuroraDSQLPool`
- `transactionWithRetry()` — an opt-in retry loop that re-runs the whole
  transaction callback on a DSQL conflict (`OC000` / `OC001` / `40001`). The
  adapter owns this loop rather than delegating to the connector's: only this
  layer sees Drizzle's wrapped `DrizzleQueryError.cause`, so the connector's
  own retry is bypassed by design.
- `migrate()` / `getMigrationStatus()` — resumable migrations that apply one DDL
  per transaction (autocommit) and record each statement in a tracking table
- `aurora-dsql-drizzle` CLI (`generate` / `transform` / `lint`) wrapping
  dsql-lint, preserving Drizzle's `--> statement-breakpoint` markers
- veterinary-app example, including the validate-then-write pattern that stands
  in for DSQL's absent foreign keys
- CI and release workflows, and the Drizzle entry in the root README

## Testing

- 133 unit tests (mocked): driver, session/transactions, retry, migrator, CLI
  argument parsing and transform/lint, SQL statement rendering
- 18 live integration tests against a real DSQL cluster: CRUD, relations, UUID
  generation, forced-conflict OCC retry and exhaustion, `CREATE INDEX ASYNC`,
  migrator apply / resume / status, and the example's reference checks and
  transaction rollback

---

*Issue #, if available:*
N/A

### DIFF
--- a/.github/workflows/node-drizzle-ci.yml
+++ b/.github/workflows/node-drizzle-ci.yml
@@ -1,0 +1,133 @@
+name: Node Drizzle CI
+
+permissions: {}
+
+on:
+  workflow_call: {}
+  workflow_dispatch:
+  push:
+    branches: [main]
+
+defaults:
+  run:
+    working-directory: node/drizzle
+
+jobs:
+  build:
+    name: Build & unit test (Node ${{ matrix.node-version }})
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [20, 22, 24]
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up Node
+        uses: actions/setup-node@v7
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Check formatting
+        run: npm run format:check
+
+      - name: Build package
+        run: npm run build
+
+      # Unit tests are fully mocked (no cluster); they only need the
+      # dsql-lint binary, which installs with the package dependencies.
+      - name: Run unit tests
+        run: npm test
+
+  create-cluster:
+    needs: build
+    uses: ./.github/workflows/dsql-cluster-create.yml
+    with:
+      workflow_name: node-drizzle
+    secrets:
+      IAM_ROLE: ${{ secrets.IAM_ROLE }}
+    permissions:
+      id-token: write # required by aws-actions/configure-aws-credentials
+
+  test-example:
+    needs: create-cluster
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # required by aws-actions/configure-aws-credentials
+    env:
+      SSL_CERT_FILE: /tmp/AmazonRootCA1.pem
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v6
+        with:
+          role-to-assume: ${{ secrets.IAM_ROLE }}
+          aws-region: ${{ needs.create-cluster.outputs.region }}
+
+      - name: Set up Node
+        uses: actions/setup-node@v7
+        with:
+          node-version: "22.x"
+
+      - name: Download Amazon Root Cert for SSL
+        run: |
+          mkdir -p ~/.postgresql
+          wget https://www.amazontrust.com/repository/AmazonRootCA1.pem -O ~/.postgresql/root.crt
+
+      - name: Install and build adapter
+        run: |
+          npm ci
+          npm run build
+
+      - name: Install and build example
+        working-directory: node/drizzle/examples/veterinary-app
+        run: |
+          npm ci
+          npm run build
+
+      - name: Run migrations
+        working-directory: node/drizzle/examples/veterinary-app
+        env:
+          CLUSTER_USER: "admin"
+          CLUSTER_ENDPOINT: ${{ needs.create-cluster.outputs.cluster-endpoint }}
+        run: npm run db:migrate
+
+      - name: Run example integration tests
+        working-directory: node/drizzle/examples/veterinary-app
+        env:
+          CLUSTER_USER: "admin"
+          CLUSTER_ENDPOINT: ${{ needs.create-cluster.outputs.cluster-endpoint }}
+        run: npm run test -- --testPathPatterns=dsql-client
+
+      # Live migrator coverage: fresh apply, idempotent re-run, status, resume
+      # after partial failure, and breakpoint-free multi-DDL rejection. Runs in
+      # an isolated `drizzle_mig_test` schema it also tears down.
+      - name: Run migrator integration tests
+        working-directory: node/drizzle/examples/veterinary-app
+        env:
+          CLUSTER_USER: "admin"
+          CLUSTER_ENDPOINT: ${{ needs.create-cluster.outputs.cluster-endpoint }}
+        run: npm run test -- --testPathPatterns=migrator
+
+      - name: Run sample
+        working-directory: node/drizzle/examples/veterinary-app
+        env:
+          CLUSTER_USER: "admin"
+          CLUSTER_ENDPOINT: ${{ needs.create-cluster.outputs.cluster-endpoint }}
+        run: npm run sample
+
+  delete-cluster:
+    if: always() && needs.create-cluster.result == 'success'
+    needs: [create-cluster, test-example]
+    uses: ./.github/workflows/dsql-cluster-delete.yml
+    with:
+      cluster-id: ${{ needs.create-cluster.outputs.cluster-id }}
+      region: ${{ needs.create-cluster.outputs.region }}
+    secrets:
+      IAM_ROLE: ${{ secrets.IAM_ROLE }}
+    permissions:
+      id-token: write # required by aws-actions/configure-aws-credentials
--- a/node/drizzle/README.md
+++ b/node/drizzle/README.md
@@ -1,0 +1,134 @@
+# Aurora DSQL adapter for Drizzle ORM
+
+[![GitHub](https://img.shields.io/badge/github-awslabs/aurora--dsql--orms-blue?logo=github)](https://github.com/awslabs/aurora-dsql-orms)
+[![npm version](https://img.shields.io/npm/v/@aws/aurora-dsql-drizzle.svg)](https://www.npmjs.com/package/@aws/aurora-dsql-drizzle)
+[![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+[![Discord chat](https://img.shields.io/discord/1435027294837276802.svg?logo=discord)](https://discord.com/invite/nEF6ksFWru)
+
+[Drizzle ORM](https://orm.drizzle.team/) support for [Amazon Aurora DSQL](https://aws.amazon.com/rds/aurora/dsql/).
+
+It rides on `drizzle-orm/node-postgres`: the [Aurora DSQL connector](https://github.com/awslabs/aurora-dsql-connectors/tree/main/node/node-postgres) is a `pg.Pool` with IAM token authentication, so there is no custom dialect — just a thin `drizzle()` factory, an opt-in OCC retry helper, a DSQL-aware migrator, and a migration CLI.
+
+## Requirements
+
+- Drizzle ORM `^0.45` and `pg` `>=8` (peer dependencies)
+- Node.js `>=20`
+- An Aurora DSQL cluster, and AWS credentials with `dsql:DbConnect` for the database role you connect as
+
+## Install
+
+```bash
+npm install @aws/aurora-dsql-drizzle drizzle-orm pg
+npm install -D drizzle-kit
+```
+
+IAM authentication and TLS are handled by the connector.
+
+## Connect
+
+```ts
+import { drizzle } from "@aws/aurora-dsql-drizzle";
+import * as schema from "./schema";
+
+const db = drizzle({
+  connection: {
+    host: process.env.CLUSTER_ENDPOINT!, // <id>.dsql.<region>.on.aws
+    region: "us-east-1", // optional; inferred from the host otherwise
+    user: "myuser", // a database role scoped to what your app needs
+    options: "-c search_path=myschema",
+  },
+  schema,
+});
+
+const owners = await db.select().from(schema.owner);
+```
+
+`user` is required — see [Using database roles and IAM authentication](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/using-database-and-iam-roles.html) for creating a role granted only the privileges your application needs. There is no default, so a connection never lands on `admin` by omission.
+
+Already have a pool? Pass it directly: `drizzle({ client: pool, schema })`, where `pool` is an `AuroraDSQLPool` (or any `pg.Pool`). `db.$client` exposes the underlying pool; call `await db.$client.end()` to close it.
+
+## Transactions with OCC retry
+
+DSQL uses optimistic concurrency control: conflicting transactions fail at COMMIT with `OC000` / `OC001` (SQLSTATE `40001`). `db.transactionWithRetry` re-runs the whole transaction on those conflicts.
+
+```ts
+import { sql } from "drizzle-orm";
+
+await db.transactionWithRetry(async (tx) => {
+  await tx.update(accounts).set({ balance: sql`balance - 100` }).where(...);
+  await tx.update(accounts).set({ balance: sql`balance + 100` }).where(...);
+});
+```
+
+The callback is re-run on every retry, so it **must be idempotent** — no side effects (emails, queue writes) that must not repeat. Keep transactions flat: a nested `tx.transaction()` fails fast with an explanatory error, so partial work is never mistaken for a committed savepoint.
+
+Retry defaults are `maxRetries` 3, `baseDelayMs` 50, `maxDelayMs` 5000 (exponential backoff with equal jitter). Pass an optional transaction config second and retry overrides third; when retries are exhausted it throws `AwsDsqlRetryExhaustedError` (the last conflict is on `.cause`).
+
+```ts
+await db.transactionWithRetry(
+  async (tx) => { await tx.insert(orders).values({ ... }); },
+  { isolationLevel: "serializable" },
+  { maxRetries: 5, onRetry: (err, attempt, max) => log.warn({ err, attempt, max }) },
+);
+```
+
+## Migrations
+
+Import `migrate` from this package. It applies one DDL statement per transaction — matching how DSQL runs DDL — and tracks each statement individually, so use it in place of the stock `drizzle-orm/node-postgres` migrator, which sends every statement in a single transaction:
+
+```ts
+import { migrate, getMigrationStatus } from "@aws/aurora-dsql-drizzle";
+
+const result = await migrate(db, { migrationsFolder: "./drizzle" });
+if (!result.success) throw new Error(result.error.message);
+```
+
+The workflow is:
+
+1. **Generate** SQL from your schema and rewrite it for DSQL:
+
+   ```bash
+   npx aurora-dsql-drizzle generate --out ./drizzle -- --config drizzle.config.ts
+   ```
+
+   This runs `drizzle-kit generate`, then rewrites each statement with [`dsql-lint`](https://github.com/awslabs/aurora-dsql-tools/tree/main/dsql-lint) — the Aurora DSQL linter and fixer — while preserving Drizzle's `--> statement-breakpoint` markers. It turns `CREATE INDEX` into `CREATE INDEX ASYNC`, rewrites `SERIAL` columns as `BIGINT … GENERATED … AS IDENTITY` (a type widening — review it), removes foreign-key constraints, and reports anything it cannot rewrite. Review and commit the result. (`transform` and `lint` subcommands run those steps on their own.)
+
+   Keep Drizzle Kit's `breakpoints: true` (the default). The adapter applies one statement per marker, so a breakpoint-free file holding more than one statement is rejected with an explanatory error rather than sent as a single multi-statement transaction.
+
+   Known limitation: the transform lints each statement separately, so a fix needing more than one statement at a time does not apply. The case to know about is `ALTER COLUMN … ADD GENERATED … AS IDENTITY`, which `dsql-lint` folds into the preceding `CREATE TABLE` when it sees both together; here it sees only the `ALTER` and reports it as unfixable. Define identity columns in the table definition, or merge the two statements by hand.
+
+2. **Apply** the committed migrations at deploy time, using the `migrate()` call above.
+
+   Each statement is applied on its own (autocommit) and then recorded in a tracking table, so a run interrupted partway resumes where it left off — recorded statements are skipped. Asynchronous DDL (`CREATE INDEX ASYNC`, `ALTER TABLE ASYNC … VALIDATE CONSTRAINT`) is awaited before the statement is recorded, so a failed background job is never reported as success. Conflicting statements are retried on DSQL's optimistic-concurrency errors. `getMigrationStatus(db, config)` reports applied vs. pending without changing anything.
+
+   One caveat on resuming: a statement and its tracking row are separate commits. If a run dies in the gap between them, the statement is applied but untracked, and because `drizzle-kit` emits `CREATE TABLE` without `IF NOT EXISTS` the re-run fails with "already exists". `migrate()` reports which statement it was so you can reconcile the tracking table by hand.
+
+## CLI
+
+```
+aurora-dsql-drizzle generate [--out <dir>] [-- <drizzle-kit args>]   Generate + transform
+aurora-dsql-drizzle transform [input] [-o output]                    Transform SQL for DSQL
+aurora-dsql-drizzle lint [input]                                      Lint SQL for DSQL
+```
+
+Exit codes: `0` clean, `1` unfixable errors remain (and the adapter's own usage errors, e.g. an unknown flag), `2` usage error propagated from dsql-lint, `3` fixed with advisories (e.g. foreign keys removed — review before applying).
+
+## Example
+
+See [examples/veterinary-app](./examples/veterinary-app/) for a complete project: schema, committed DSQL migration, `db:migrate` script, and integration tests against a live cluster.
+
+## Resources
+
+- [Amazon Aurora DSQL documentation](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/what-is-aurora-dsql.html)
+- [Unsupported PostgreSQL features in DSQL](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-unsupported-features.html)
+- [Aurora DSQL connector for node-postgres](https://github.com/awslabs/aurora-dsql-connectors/tree/main/node/node-postgres)
+- [Drizzle ORM documentation](https://orm.drizzle.team/docs)
+- [dsql-lint](https://github.com/awslabs/aurora-dsql-tools/tree/main/dsql-lint)
+
+## Security
+
+See [CONTRIBUTING](../../CONTRIBUTING.md#security-issue-notifications) for more information.
+
+## License
+
+Apache-2.0
--- a/node/drizzle/examples/veterinary-app/README.md
+++ b/node/drizzle/examples/veterinary-app/README.md
@@ -1,0 +1,52 @@
+# Veterinary app (Drizzle + Aurora DSQL)
+
+A small veterinary clinic schema demonstrating [@aws/aurora-dsql-drizzle](../../): IAM-authenticated connections, DSQL-ready migrations, and relational queries.
+
+## Layout
+
+- `src/schema.ts` — Drizzle tables. UUID primary keys, and relationships declared at the ORM level with `relations()` rather than `references()`, so the generated SQL carries no foreign-key constraints
+- `drizzle/` — the committed migration, generated with `drizzle-kit` and rewritten for DSQL (`CREATE INDEX ASYNC`, `USING btree` removed)
+- `src/dsql-client.ts` — builds the `drizzle()` database
+- `src/migrate.ts` — applies migrations via the adapter's `migrate()`
+- `src/example.ts` — populates and verifies data
+
+## Run
+
+```bash
+export CLUSTER_ENDPOINT=<your-cluster-id>.dsql.<region>.on.aws
+export CLUSTER_USER=myuser   # required; the database role to connect as
+
+npm install
+npm run db:migrate   # applies drizzle/ with the adapter's migrate()
+npm run sample       # populate + verify
+npm test             # integration tests (needs the cluster)
+```
+
+Credentials come from the default AWS credential chain, and need `dsql:DbConnect` for the role named in `CLUSTER_USER`.
+
+### Setting up the database role
+
+Connect as `admin` once to create an application role scoped to just this app's schema, plus the schema `migrate()` tracks applied statements in. See [Using database roles and IAM authentication](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/using-database-and-iam-roles.html):
+
+```sql
+CREATE SCHEMA myschema;
+CREATE SCHEMA drizzle;   -- migrate()'s tracking schema
+CREATE ROLE myuser WITH LOGIN;
+AWS IAM GRANT myuser TO 'arn:aws:iam::<account-id>:role/<your-iam-role>';
+GRANT USAGE, CREATE ON SCHEMA myschema TO myuser;
+GRANT USAGE, CREATE ON SCHEMA drizzle TO myuser;
+```
+
+`CLUSTER_USER=myuser` makes the app connect with `search_path=myschema`. There is no default, so the app never lands on `admin` and the shared `public` schema by omission.
+
+Run `db:migrate` and the app as the same `CLUSTER_USER`, and don't change it afterwards. The schema is derived from it, while `migrate()` tracks applied statements globally — so migrating as `admin` puts the tables in `public`, and a later run as `myuser` finds every statement already recorded and leaves `myschema` empty. One role also keeps the grants above sufficient: `myuser` owns the tables it creates, so nothing needs granting after a migration.
+
+## Regenerating the migration
+
+After changing `src/schema.ts`:
+
+```bash
+npm run db:generate   # aurora-dsql-drizzle generate: drizzle-kit generate + dsql-lint --fix
+```
+
+This is the adapter's own CLI wrapping both steps, so plain `drizzle-kit generate` isn't needed — the second step is what rewrites the generated SQL for DSQL. Review and commit the result.
--- a/node/drizzle/examples/veterinary-app/package.json
+++ b/node/drizzle/examples/veterinary-app/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "veterinary-app",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Veterinary clinic example using Drizzle ORM with Aurora DSQL",
+  "main": "dist/index.js",
+  "scripts": {
+    "db:generate": "aurora-dsql-drizzle generate --out ./drizzle -- --config drizzle.config.ts",
+    "db:migrate": "tsx src/migrate.ts",
+    "format": "prettier --write .",
+    "format:check": "prettier --check .",
+    "build": "tsc",
+    "presample": "npm run build",
+    "sample": "tsx src/index.ts",
+    "pretest": "npm run build",
+    "test": "NODE_OPTIONS='--experimental-vm-modules' jest"
+  },
+  "devDependencies": {
+    "@types/jest": "^30.0.0",
+    "@types/node": "^25.2.0",
+    "@types/pg": "^8.15.0",
+    "drizzle-kit": "^0.31.0",
+    "prettier": "4.0.0-alpha.13",
+    "ts-jest": "^29.4.1",
+    "tsx": "^4.20.3",
+    "typescript": "^5.8.3"
+  },
+  "dependencies": {
+    "@aws/aurora-dsql-drizzle": "file:../..",
+    "drizzle-orm": "^0.45.2",
+    "pg": "^8.16.0"
+  }
+}
--- a/node/drizzle/examples/veterinary-app/src/schema.ts
+++ b/node/drizzle/examples/veterinary-app/src/schema.ts
@@ -1,0 +1,85 @@
+import { relations, sql } from "drizzle-orm";
+import {
+  date,
+  index,
+  pgTable,
+  primaryKey,
+  uuid,
+  varchar,
+} from "drizzle-orm/pg-core";
+
+// Primary keys are UUIDs defaulted with gen_random_uuid(): random keys spread
+// writes across DSQL's distributed storage instead of contending on one end of
+// a monotonic sequence. Integer auto-increment is available too — via identity
+// columns (`GENERATED ... AS IDENTITY`), which is what dsql-lint rewrites the
+// `SERIAL` keyword into.
+//
+// Relationships are declared with `relations()` at the ORM level rather than
+// `references()`, so the generated SQL carries no foreign-key constraints.
+
+export const owner = pgTable("owner", {
+  id: uuid("id")
+    .primaryKey()
+    .default(sql`gen_random_uuid()`),
+  name: varchar("name", { length: 30 }).notNull(),
+  city: varchar("city", { length: 80 }).notNull(),
+  telephone: varchar("telephone", { length: 20 }),
+});
+
+export const pet = pgTable(
+  "pet",
+  {
+    id: uuid("id")
+      .primaryKey()
+      .default(sql`gen_random_uuid()`),
+    name: varchar("name", { length: 30 }).notNull(),
+    birthDate: date("birth_date", { mode: "date" }).notNull(),
+    ownerId: uuid("owner_id"),
+  },
+  (table) => [index("pet_owner_id_idx").on(table.ownerId)],
+);
+
+export const specialty = pgTable("specialty", {
+  name: varchar("name", { length: 80 }).primaryKey(),
+});
+
+export const vet = pgTable("vet", {
+  id: uuid("id")
+    .primaryKey()
+    .default(sql`gen_random_uuid()`),
+  name: varchar("name", { length: 30 }).notNull(),
+});
+
+// Join table for the many-to-many between vets and specialties.
+export const specialtyToVet = pgTable(
+  "specialty_to_vet",
+  {
+    specialtyName: varchar("specialty_name", { length: 80 }).notNull(),
+    vetId: uuid("vet_id").notNull(),
+  },
+  (table) => [primaryKey({ columns: [table.specialtyName, table.vetId] })],
+);
+
+export const ownerRelations = relations(owner, ({ many }) => ({
+  pets: many(pet),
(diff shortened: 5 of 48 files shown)


node/drizzle/src/retry.ts
```diff
@@ -0,0 +1,74 @@
...
+        "adapter's drizzle() (backed by an AuroraDSQLPool with OCC retry).",
+    );
+  }
+
+  // Reuse the schema so the relational query API (db.query.*) works inside the
+  // retried transaction, just as it does on the top-level db.
+  const schema = (db as unknown as { _?: { fullSchema?: TSchema } })._
+    ?.fullSchema;
+
+  return (pool as AuroraDSQLPool).transaction((client: PoolClient) => {
```
amaksimo: `AuroraDSQLPool.transaction()` retries only when connector 0.1.9's `isOCCError` sees a top-level `.code`, but Drizzle wraps pg failures in `DrizzleQueryError` at `error.cause.code`. I reproduced a `tx.execute()` throwing `40001` with `maxRetries: 2`: one connection and one query attempt. Please unwrap/rethrow the pg cause (or update connector handling) and add a regression through a real Drizzle query.

node/drizzle/src/retry.ts
```diff
@@ -0,0 +1,74 @@
...
+    const txDb = nodePgDrizzle(
```
amaksimo: This rebuilds Drizzle with only `schema`, dropping `casing`, logger, and cache configuration. With `casing: "snake_case"`, I reproduced the outer query emitting `"camel_name"` while the same query in this callback emitted `"camelName"`, which can target a nonexistent column. Please preserve the original dialect/session configuration.

node/drizzle/src/retry.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @param db A Drizzle instance created by this adapter's `drizzle()`.
+ * @param transaction The (idempotent) transaction callback.
+ * @param config Optional per-call OCC retry overrides.
+ */
+export async function transactionWithRetry<
+  TSchema extends Record<string, unknown> = Record<string, never>,
+  T = unknown,
+>(
+  db: NodePgDatabase<TSchema> & { $client: unknown },
+  transaction: (tx: NodePgDatabase<TSchema>) => Promise<T>,
```
amaksimo: The callback is typed and constructed as `NodePgDatabase`, not a Drizzle transaction. Calling the familiar `tx.transaction()` starts and commits another transaction on the connector-owned client; I reproduced `OUTER_BEGIN, begin, select 1, commit, OUTER_COMMIT`, so the inner `commit` ends the outer transaction early. Please expose a true transaction shape and reject nested transactions clearly, as drizzle-team/drizzle-orm#5538 does.

node/drizzle/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "engines": {
+    "node": ">=20"
+  },
+  "peerDependencies": {
+    "drizzle-orm": "^0.45",
+    "pg": ">=8"
+  },
+  "dependencies": {
+    "@aws/aurora-dsql-node-postgres-connector": "^0.1.9",
+    "@aws/dsql-lint": "^0.1"
```
amaksimo: `^0.1` currently locks 0.1.3. I ran it: it rejects now-supported `DROP COLUMN`, `DROP NOT NULL`, `SET/DROP DEFAULT`, and cannot parse `ALTER TABLE ASYNC ... VALIDATE CONSTRAINT`. Version 0.2.14 fixes those, but still rejects the documented `ALTER COLUMN ... ADD GENERATED ... AS IDENTITY` form. Please depend on a dsql-lint release aligned with the [current ALTER TABLE surface](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/alter-table-syntax-support.html).

node/drizzle/src/migrator.ts
```diff
@@ -0,0 +1,384 @@
...
+          ? `${stmt.slice(0, 500)}... [truncated, ${stmt.length - 500} more chars]`
+          : stmt,
+    },
+  });
+
+  for (const migration of migrations) {
+    let migrationHadPendingStatements = false;
+
+    for (let stmtIdx = 0; stmtIdx < migration.sql.length; stmtIdx++) {
+      const stmtKey = `${migration.hash}:${stmtIdx}`;
```
amaksimo: `migration.hash` is Drizzle's SHA-256 of the entire SQL file, yet it is part of every statement's identity. Even adding a trailing newline changes all keys, replays previously applied DDL, and makes the changed-statement warning unreachable for real `readMigrationFiles()` output. Please use stable journal identity plus statement index, keeping the statement hash only for mutation detection.

node/drizzle/src/cli/transform.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * Transform a whole migration file for DSQL. Drizzle splits statements with
+ * `--> statement-breakpoint`; each statement is transformed independently and
+ * the markers are re-inserted so the runtime migrator can still split the file.
+ * Statements that dsql-lint removes entirely (e.g. foreign keys) are dropped.
+ */
+export function transformMigrationFile(sql: string): FileTransformResult {
+  if (!sql.includes(BREAKPOINT)) {
```
amaksimo, line 75: Drizzle Kit supports `breakpoints: false`; without a marker, this leaves a multi-statement file as one chunk, and the migrator later sends multiple DDLs in one `db.execute()`. [DSQL transactions can contain only one DDL](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-ddl.html). Please split safely or reject breakpoint-free multi-DDL migrations, and add an execution regression.

node/drizzle/src/migrator.ts
```diff
@@ -0,0 +1,384 @@
...
+      // Empty chunk (e.g. trailing statement-breakpoint) — still track it so
+      // statement indices stay aligned with the file on the next run.
+      if (!stmt) {
+        await recordStatement(migration, stmtIdx, currentStmtHash);
+        appliedStatementsCount++;
+        continue;
+      }
+
+      try {
+        await db.execute(sql.raw(stmt));
```
amaksimo: [`CREATE INDEX ASYNC` returns immediately with a job ID](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-create-index-async.html), but this discards the result and immediately records the statement complete. If the background job later fails, `migrate()` reports success and future runs skip the missing index. Please wait or poll for terminal success before recording completion.

node/drizzle/tests/retry.test.ts
```diff
@@ -0,0 +1,93 @@
+import { pgTable, text } from "drizzle-orm/pg-core";
+import { transactionWithRetry, isOCCError } from "../src/retry";
+
+describe("transactionWithRetry", () => {
+  test("delegates to the connector pool's transaction() and returns its result", async () => {
```
amaksimo: These tests only verify delegation to a mock whose `transaction()` invokes the callback once; neither the connector retry loop nor Drizzle error wrapping is exercised. The live test also covers only success. Since OCC retry is a core feature, please add a real Drizzle query over a fake `PoolClient` returning `40001` and assert attempts/limits, plus a live coordinated conflict proving the callback reruns and commits. The Hibernate `LockingTest` in this repo already demonstrates a deterministic latch-based `OC000` pattern.

node/drizzle/tests/migrator.test.ts
```diff
@@ -0,0 +1,251 @@
...
+    executed.push(text);
+    if (/^\s*SELECT\s+migration_hash/i.test(text)) {
+      return { rows: appliedRows };
+    }
+    return { rows: [] };
+  });
+  return { db: { execute } as never, executed };
+}
+
+const MIGRATION: MigrationMeta = {
```
amaksimo, line 78: All migrator tests call `migrateInternal()` with hand-built `MigrationMeta` and mocked `execute()`. This bypasses the public `migrate()` path, `readMigrationFiles()`, real journal/hash behavior, and database state. Please add public/live tests using an actual migration directory that cover first apply, rerun, status, partial failure/resume, and resulting schema. The Tortoise Aerich suite in this repo exercises generated incremental migrations against live DSQL in this way.

node/drizzle/tests/transform.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    expect(worseExitCode(0, 3)).toBe(3);
+    expect(worseExitCode(3, 0)).toBe(3);
+  });
+
+  test("clean stays clean", () => {
+    expect(worseExitCode(0, 0)).toBe(0);
+  });
+});
+
+describe("transformMigrationFile (Drizzle statement-breakpoint)", () => {
```
amaksimo, line 181: This covers marker-separated input and a single statement without markers, but not a breakpoint-free multi-statement file. Since migration safety depends on statement boundaries, please add multi-statement cases with semicolons in quoted strings, comments, and dollar quotes, plus live multiple-DDL execution. Tortoise has dedicated `split_sql` edge-case tests and a live multiple-DDL integration test that provide a good repository-local model.

.github/workflows/node-drizzle-ci.yml
```diff
@@ -0,0 +1,123 @@
...
+          npm run build
+
+      - name: Run migrations
+        working-directory: node/drizzle/examples/veterinary-app
+        env:
+          CLUSTER_USER: "admin"
+          CLUSTER_ENDPOINT: ${{ needs.create-cluster.outputs.cluster-endpoint }}
+        run: npm run db:migrate
+
+      - name: Run example integration tests
```
amaksimo, line 99: Unlike the Tortoise, Django, SQLAlchemy, and Hibernate workflows, this runs no dedicated adapter integration suite; it applies one migration once and then runs example CRUD. Please add live adapter coverage for connector/config propagation, transaction commit/rollback/OCC, migration rerun/status/resume, multiple-DDL handling, and async-index terminal state. Those are the DSQL-specific guarantees this package owns.

amaksimo: To add, the existing OCC tests are not sufficient. I left this comment previously, they do not force an OCC. Please structure the test in a way that will guarantee there will be an OCC conflict.

amaksimo:
We have two open PRs for upstreaming DSQL support into Drizzle: [drizzle-orm#5538](https://github.com/drizzle-team/drizzle-orm/pull/5538) and [drizzle-orm#5310](https://github.com/drizzle-team/drizzle-orm/pull/5310). #5538 is the closest implementation reference for this adapter, so why does this PR differ from it in several core behaviors?

Unless these differences are required by an external-adapter constraint, please align this implementation with the upstream shape before merging:

1. Use the same retry classification semantics, including recursively inspecting wrapped `cause` errors.
2. Preserve the existing Drizzle schema, relations, logger/casing, dialect, and session options when creating retry transaction sessions.
3. Keep the transaction callback idiomatic for Drizzle by exposing the transaction object and explicitly rejecting unsupported nested transactions.
4. Match the upstream testing baseline: induce real concurrent OCC conflicts, verify retry exhaustion, and cover live migration application, reruns, multiple migrations/statements, partial failure, tracking state, and resulting schema state.

The implementations do not need to be byte-for-byte identical, but their public API, connector behavior, retry semantics, and test contract should remain close. Please document any intentional divergence. The migration identity and asynchronous-index completion issues raised elsewhere appear in the upstream work as well; those should be corrected in both implementations rather than copied here.

amaksimo:
> We have two open PRs for upstreaming DSQL support into Drizzle: [drizzle-orm#5538](https://github.com/drizzle-team/drizzle-orm/pull/5538) and [drizzle-orm#5310](https://github.com/drizzle-team/drizzle-orm/pull/5310). #5538 is the closest implementation reference for this adapter, so why does this PR differ from it in several core behaviors?
> 
> Unless these differences are required by an external-adapter constraint, please align this implementation with the upstream shape before merging:
> 
> 1. Use the same retry classification semantics, including recursively inspecting wrapped `cause` errors.
> 2. Preserve the existing Drizzle schema, relations, logger/casing, dialect, and session options when creating retry transaction sessions.
> 3. Keep the transaction callback idiomatic for Drizzle by exposing the transaction object and explicitly rejecting unsupported nested transactions.
> 4. Match the upstream testing baseline: induce real concurrent OCC conflicts, verify retry exhaustion, and cover live migration application, reruns, multiple migrations/statements, partial failure, tracking state, and resulting schema state.
> 
> The implementations do not need to be byte-for-byte identical, but their public API, connector behavior, retry semantics, and test contract should remain close. Please document any intentional divergence. The migration identity and asynchronous-index completion issues raised elsewhere appear in the upstream work as well; those should be corrected in both implementations rather than copied here.

To be clear, the final API we give to customers should match what we have in the upstream PR, so when we upstream, customers can easily change.

That part is very important ^

praba2210:
Thanks — #5538 is the right reference, so we sorted our divergences from it into **converging** (gaps we'll close) and **intentional** (constrained by shipping on stable Drizzle). Verified against connector `0.1.9` / `drizzle-orm@0.45.2`.

The root cause is a single early decision: we delegated OCC retry to the connector's `transaction()` instead of owning it at the Drizzle layer. The connector passes a raw `PoolClient` and classifies only a top-level `.code`, so we rebuilt Drizzle per attempt and never saw Drizzle's wrapped errors — which is why the retry issues cluster. Owning the loop at the Drizzle layer resolves them together.

### Converging — gaps

- **OCC retry never fires:** the connector's `isOCCError` reads only a top-level `.code`, but Drizzle wraps pg errors at `.cause`. We'll recursively unwrap `cause`, matching `isDsqlRetryableError`, with a regression through a real Drizzle query.
- **Retry drops session config:** we rebuild with only `schema` — we'll preserve casing/logger/cache/dialect.
- **Callback isn't a real transaction:** it hands back a `NodePgDatabase`, so nested `tx.transaction()` commits the outer. We'll expose a real transaction and reject nesting.

These three are fixed together by moving the retry loop into the adapter (wrapping Drizzle's own `db.transaction`), reusing the connector for IAM/pooling and its OCC classifier + backoff — the structure #5538 uses.

Also converging:

- Migration identity (journal identity + statement index; statement hash for mutation detection only — same bug exists in #5538)
- `CREATE INDEX ASYNC` completion polling
- Breakpoint-free multi-DDL handling
- The `dsql-lint` bump
- Test baseline: induced OCC + retry exhaustion; live migration apply/rerun/status/resume/multi-statement/schema

### Intentional — one, version-driven

#5538 targets `drizzle-orm@1.0.0-beta.23` and extends `PgAsyncDatabase`/`PgAsyncSession`, which don't exist in stable (`0.45.2`). So the internal type hierarchy can't be mirrored without moving the package onto the unreleased 1.0 beta.

We'll match the **customer-facing call surface** on stable — `db.transactionWithRetry(fn, txConfig, retryConfig)`, `AwsDsqlRetryExhaustedError`, `onRetry`, matching config keys, nested-tx rejection — so switching to upstream is a no-op for customer code. Since #5538 lands in Drizzle 1.0, customers cross a major bump regardless.

**Proposal:** match the call surface on stable now and document the internal-type divergence. Or, if you'd prefer we track 1.0-beta to mirror the internals too, we'll do that — which do you want?

praba2210:
Summary of what's changed since the review (verified against connector `0.1.9` / `drizzle-orm@0.45.2`):

**Retry** — moved the loop to the adapter (wrapping Drizzle's own `db.transaction`):
- `isDsqlRetryableError` walks the `.cause` chain and delegates classification to the connector's `isOCCError` — single source of truth for the retryable codes. Regression over a wrapped `40001`.
- Reuses the same dialect (incl. `casing`), schema, logger, cache per attempt — `snake_case` mismatch gone.
- Callback gets a real `AwsDsqlTransaction`; nested `tx.transaction()` throws a clear "no savepoints" error instead of committing the outer.

**Migrator:**
- Identity is `(folder_millis, statement_index)`; statement hash kept only for mutation detection, so edits no longer replay applied DDL.
- `CREATE INDEX ASYNC` blocks on `sys.wait_for_job` and treats anything other than `succeeded = true` (incl. null/timeout) as failure.
- Breakpoint-free multi-DDL rejected at transform and runtime.

**dsql-lint** bumped to `^0.2.14`; the remaining `ALTER COLUMN … ADD GENERATED … AS IDENTITY` gap is documented.

**Tests & CI:** fake-`PoolClient` retry/exhaustion + a live OCC test (8 txns on one counter, final must equal N); live migrator suite via public `migrate()` (apply/re-run/status/resume/multi-DDL/schema); transform boundary + multi-statement cases. CI now runs dedicated migrator and client integration jobs.

**Upstream / retry-in-connector:** call surface matches #5538 — `db.transactionWithRetry(callback, transactionConfig?, retryConfig?)`, `AwsDsqlRetryExhaustedError`, `onRetry` — so migrating is a one-line import swap. Classification delegates to `isOCCError`; the loop stays adapter-side (only layer that sees Drizzle's wrapped `.cause` and can return a real `tx`; connector `transaction()` yields a raw `PoolClient`). Sharing the backoff would need the connector to export its helper — a follow-up.

**Intentional divergence (one):** #5538 targets `drizzle-orm@1.0-beta` (`PgAsyncSession`/`PgAsyncDatabase`), absent on stable `0.45`. We ship on stable and mirror the call surface, so only the internal type hierarchy differs.

node/drizzle/examples/veterinary-app/src/schema.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+  date,
+  index,
+  pgTable,
+  primaryKey,
+  uuid,
+  varchar,
+} from "drizzle-orm/pg-core";
+
+// Aurora DSQL does not support foreign keys or SERIAL, so primary keys are
```
amaksimo: SERIAL is supported

node/drizzle/README.md
```diff
@@ -0,0 +1,128 @@
...
+await db.transactionWithRetry(
+  async (tx) => { await tx.insert(orders).values({ ... }); },
+  { isolationLevel: "serializable" },
+  { maxRetries: 5, onRetry: (err, attempt, max) => log.warn({ err, attempt, max }) },
+);
+```
+
+## Migrations
+
+DSQL applies one DDL statement per transaction and expects DSQL-flavored DDL — asynchronous indexes, no foreign keys, no `SERIAL`. The workflow is:
```
amaksimo: SERIAL is fine?

node/drizzle/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+1. **Generate** SQL from your schema and rewrite it for DSQL:
+
+   ```bash
+   npx aurora-dsql-drizzle generate --out ./drizzle -- --config drizzle.config.ts
+   ```
+
+   This runs `drizzle-kit generate`, then `dsql-lint --fix` over each statement — turning `CREATE INDEX` into `CREATE INDEX ASYNC`, dropping foreign keys, etc. — while preserving Drizzle's `--> statement-breakpoint` markers. Review and commit the result. (`transform` and `lint` subcommands are available to run those steps on their own.)
```
amaksimo: We can probably link to dsql-lint here, assume customer does not know what that is.

In general, we should probably remove language from README that is like "DSQL does not support XYZ" as we are adding support for that fairly quickly, and our public facing docs will drift over time.

Please remove this and other similar "DSQL does not support" and instead write things like "the drizzle adapter breaks up txns into one DDL per txn" rather than DSQL does not support... the focus on what the adapter does instead.

node/drizzle/README.md
```diff
@@ -0,0 +1,128 @@
...
+- An Aurora DSQL cluster and AWS credentials with `dsql:DbConnect` / `dsql:DbConnectAdmin`
+
+## Install
+
+```bash
+npm install @aws/aurora-dsql-drizzle drizzle-orm pg
+npm install -D drizzle-kit
+```
+
+IAM authentication and TLS are handled by the connector — you provide the cluster host, not a password.
```
amaksimo: — you provide the cluster host, not a password.

We can leave that out lol

node/drizzle/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```ts
+import { drizzle } from "@aws/aurora-dsql-drizzle";
+import * as schema from "./schema";
+
+const db = drizzle({
+  connection: {
+    host: process.env.CLUSTER_ENDPOINT!, // <id>.dsql.<region>.on.aws
+    region: "us-east-1", // optional; inferred from the host otherwise
+    // user defaults to "admin"; use a database role + search_path otherwise:
```
amaksimo: The best practice is to NOT use an admin role. As such, we should NOT default to admin.

The user should be directed to create a minimally scoped role as is the IAM guideline.

node/drizzle/README.md
```diff
@@ -0,0 +1,128 @@
...
+```ts
+import { sql } from "drizzle-orm";
+
+await db.transactionWithRetry(async (tx) => {
+  await tx.update(accounts).set({ balance: sql`balance - 100` }).where(...);
+  await tx.update(accounts).set({ balance: sql`balance + 100` }).where(...);
+});
+```
+
+The callback is re-run on every retry, so it **must be idempotent** — no side effects (emails, queue writes) that must not repeat. DSQL has no savepoints, so a nested `tx.transaction()` throws a clear error instead of silently committing the outer transaction.
```
amaksimo: refactor "DSQL has no savepoints, so a nested `tx.transaction()` throws a clear error instead of silently committing the outer transaction." the negative sentence to a more positive one.

node/drizzle/README.md
```diff
@@ -0,0 +1,128 @@
...
+   ```ts
+   import { migrate } from "@aws/aurora-dsql-drizzle";
+
+   const result = await migrate(db, { migrationsFolder: "./drizzle" });
+   if (!result.success) throw new Error(result.error.message);
+   ```
+
+   `migrate()` applies each statement individually (autocommit) and records it in a tracking table, so a run interrupted partway can be safely re-run — already-applied statements are skipped. For `CREATE INDEX ASYNC` it waits for the background build to finish before recording the statement, so a failed build isn't reported as success. `getMigrationStatus(db, config)` reports applied vs. pending without changing anything.
+
+The stock `drizzle-orm/node-postgres` `migrate()` wraps all statements in one transaction, which DSQL doesn't allow — use this `migrate()` instead.
```
amaksimo: Again negative wording. Remove that please.

Let's also push this up higher. We are explicitly saying that migrate() stock is not supported! We need to show an example of what to do here and what is supported to make it clear for customers.

Also can we hide stock migrate from the adapter to avoid confusion?

node/drizzle/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "engines": {
+    "node": ">=20"
+  },
+  "peerDependencies": {
+    "drizzle-orm": "^0.45",
+    "pg": ">=8"
+  },
+  "dependencies": {
+    "@aws/aurora-dsql-node-postgres-connector": "^0.1.9",
+    "@aws/dsql-lint": "^0.2.14"
```
amaksimo: We should NEVER pin exact version of dsql-lint, make it always grab latest, it is constantly evolving!

node/drizzle/package.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "engines": {
+    "node": ">=20"
+  },
+  "peerDependencies": {
+    "drizzle-orm": "^0.45",
+    "pg": ">=8"
+  },
+  "dependencies": {
+    "@aws/aurora-dsql-node-postgres-connector": "^0.1.9",
```
amaksimo: do not pin to fixed minor version.

node/drizzle/examples/veterinary-app/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@types/pg": "^8.15.0",
+    "drizzle-kit": "^0.31.0",
+    "prettier": "4.0.0-alpha.13",
+    "ts-jest": "^29.4.1",
+    "tsx": "^4.20.3",
+    "typescript": "^5.8.3"
+  },
+  "dependencies": {
+    "@aws/aurora-dsql-drizzle": "file:../..",
+    "@aws/aurora-dsql-node-postgres-connector": "^0.1.9",
```
amaksimo: do not pin?

also we need this dep at all? should be within drizzle adapter, no?

node/drizzle/examples/veterinary-app/README.md
```diff
@@ -0,0 +1,35 @@
+# Veterinary app (Drizzle + Aurora DSQL)
+
+A small veterinary clinic schema demonstrating [@aws/aurora-dsql-drizzle](../../): IAM-authenticated connections, DSQL-ready migrations, and relational queries.
+
+## Layout
+
+- `src/schema.ts` — Drizzle tables (UUID keys, ORM-level relations; no foreign keys, as DSQL does not support them)
```
amaksimo: please avoid writing things like DSQL does not support xyz. This will change and this will go out of date.

node/drizzle/examples/veterinary-app/README.md
```diff
@@ -0,0 +1,35 @@
...
+- `drizzle/` — the committed migration, generated with `drizzle-kit` and rewritten for DSQL (`CREATE INDEX ASYNC`, `USING btree` removed)
+- `src/dsql-client.ts` — builds the `drizzle()` database
+- `src/migrate.ts` — applies migrations via the adapter's `migrate()`
+- `src/example.ts` — populates and verifies data
+
+## Run
+
+```bash
+export CLUSTER_ENDPOINT=<your-cluster-id>.dsql.us-east-1.on.aws
+export CLUSTER_USER=admin   # optional; defaults to admin
```
amaksimo: let's not default to admin please. This is bad practice as it leaves wide open permission.

node/drizzle/examples/veterinary-app/README.md
```diff
@@ -0,0 +1,35 @@
...
+- `src/schema.ts` — Drizzle tables (UUID keys, ORM-level relations; no foreign keys, as DSQL does not support them)
+- `drizzle/` — the committed migration, generated with `drizzle-kit` and rewritten for DSQL (`CREATE INDEX ASYNC`, `USING btree` removed)
+- `src/dsql-client.ts` — builds the `drizzle()` database
+- `src/migrate.ts` — applies migrations via the adapter's `migrate()`
+- `src/example.ts` — populates and verifies data
+
+## Run
+
+```bash
+export CLUSTER_ENDPOINT=<your-cluster-id>.dsql.us-east-1.on.aws
```
amaksimo: <region>?

node/drizzle/examples/veterinary-app/README.md
```diff
@@ -0,0 +1,35 @@
...
+export CLUSTER_USER=admin   # optional; defaults to admin
+
+npm install
+npm run db:migrate   # apply the committed migration
```
amaksimo: do we want them to run our custom migrate instead?

node/drizzle/examples/veterinary-app/README.md
```diff
@@ -0,0 +1,35 @@
...
+npm run sample       # populate + verify
+npm test             # integration tests (needs the cluster)
+```
+
+Uses the default AWS credential chain — configure credentials with `dsql:DbConnectAdmin` for the cluster.
```
amaksimo: We should not use admin.

node/drizzle/examples/veterinary-app/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Regenerating the migration
+
+After changing `src/schema.ts`:
+
+```bash
+npm run db:generate   # drizzle-kit generate + dsql-lint --fix into ./drizzle
```
amaksimo: Why do we use this path if we can use built-in command that does drizzle-kit generate + dsql-lint after?

Aside: we can modify drizzle such that it will not generate SQL that is not compat with dsql-lint and then we can avoid this (not sure)?

node/drizzle/src/cli/index.ts
```diff
@@ -0,0 +1,348 @@
...
+  transformMigrationFile,
+  worseExitCode,
+} from "./transform";
+import type { DsqlLintJsonOutput } from "./dsql-lint";
+
+const HELP = `
+Aurora DSQL Drizzle Tools
+
+Usage:
+  aurora-dsql-drizzle generate [-- <drizzle-kit args>]   Generate + transform migrations
```
amaksimo, line 16: This should be clearly stated as the golden/suggested path.

node/drizzle/src/cli/index.ts
```diff
@@ -0,0 +1,348 @@
...
+    }
+  }
+}
+
+/**
+ * Walk the dsql-lint JSON output and emit human-readable lines on stderr.
+ * Labels match dsql-lint's own text mode. See `severityFor` for the
+ * (small, open) label set.
+ */
+function reportDsqlLintDiagnostics(output: DsqlLintJsonOutput): void {
```
amaksimo, line 61: We need testing for all sorts of cases for this to ensure that we get the correct shape we are expecting!

node/drizzle/examples/veterinary-app/src/schema.ts
amaksimo: this comment is now incomplete — DSQL added identity columns + sequences (Feb 2026), and the migrator itself already uses `GENERATED ALWAYS AS IDENTITY`. the `SERIAL` keyword is still rejected (dsql-lint rewrites it), but integer auto-increment PKs *are* supported now via identity columns. i'd reword to something like "no `SERIAL` keyword — use identity columns or UUIDs" and keep the note that UUIDs are the recommended PK for distributed workloads. same wording in README.md:75.

node/drizzle/README.md
```diff
@@ -0,0 +1,128 @@
...
+2. **Apply** the committed migrations at deploy time:
+
+   ```ts
+   import { migrate } from "@aws/aurora-dsql-drizzle";
+
+   const result = await migrate(db, { migrationsFolder: "./drizzle" });
+   if (!result.success) throw new Error(result.error.message);
+   ```
+
+   `migrate()` applies each statement individually (autocommit) and records it in a tracking table, so a run interrupted partway can be safely re-run — already-applied statements are skipped. For `CREATE INDEX ASYNC` it waits for the background build to finish before recording the statement, so a failed build isn't reported as success. `getMigrationStatus(db, config)` reports applied vs. pending without changing anything.
```
amaksimo: "can be safely re-run" oversells it a bit — drizzle-kit emits `CREATE TABLE` without `IF NOT EXISTS`, so if a run crashes between applying a statement and recording it, the statement is applied-but-untracked and the re-run fails with "already exists" (needs manual reconciliation). the CREATE INDEX ASYNC handling you added is a nice improvement; i'd just add the applied-but-untracked caveat here to match the honesty of the code docstring.

amaksimo (commented):
the rework looks great — subclassing and exposing `db.transactionWithRetry()` as a method, throwing a clear error on nested `tx.transaction()`, and adding a live concurrent-contention retry test closes basically everything from my first pass. the cause-chain classifier (`isDsqlRetryableError` walking `.cause`) is the right call too — a top-level `isOCCError` check would miss conflicts raised inside a query. no correctness bugs found.

three small things left, none blocking (two inline). also:

**compat-canary** — drizzle isn't registered in `.github/workflows/compat-canary.yml` (not in this diff, so noting here). every other adapter (hibernate/django/sqlalchemy/tortoise/prisma) has a weekly "vs latest" job; the PR wires up ci-gate/pre-commit/dependabot but misses this one, so we get no early warning when `drizzle-orm` ships a breaking release. can we add a `drizzle` job mirroring the `prisma` one?

**retry ownership (fyi, not a change request)** — the adapter now owns its retry loop (`retry.ts`, defaults 50ms→5000ms) instead of delegating to the connector's `.transaction()`. reasonable — it's what lets you walk the DrizzleQueryError cause chain — but it does mean the connector's own OCC retry is now bypassed, so there are two retry impls in the stack. worth a sentence in the description so it's intentional on the record.

node/drizzle/src/migrator.ts
```diff
@@ -0,0 +1,462 @@
...
+      try {
+        const execResult = (await db.execute(sql.raw(stmt))) as QueryResult<
+          Record<string, unknown>
+        >;
+
+        // For `CREATE INDEX ASYNC`, DSQL returns immediately with a job_id
+        // while the build runs in the background. Block until the job
+        // completes so a follow-up migration (or the caller) can rely on
+        // the index actually existing.
+        if (isCreateIndexAsync(stmt)) {
```
amaksimo: Could we make this wait on every asynchronous DDL job rather than only `CREATE INDEX ASYNC`? DSQL also returns a `job_id` for `ALTER TABLE ASYNC ... VALIDATE CONSTRAINT`, and `@aws/dsql-lint@0.2.14` accepts that supported syntax. Today that migration bypasses this branch and is recorded as applied before validation finishes or fails. Please detect and wait for returned DDL job IDs generically (and fail when a known async DDL returns no `job_id`), with a validation-job test.

node/drizzle/src/migrator.ts
```diff
@@ -0,0 +1,462 @@
...
+            "only one DDL per transaction, and this migrator applies one " +
+            "statement per chunk. Regenerate with Drizzle Kit `breakpoints: true` " +
+            "(the default) so each statement is separated by a " +
+            "`--> statement-breakpoint` marker. " +
+            "See https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-ddl.html",
+        );
+      }
+
+      try {
+        const execResult = (await db.execute(sql.raw(stmt))) as QueryResult<
```
amaksimo: This plain `db.execute()` bypasses OCC retry. Connector 0.1.9 only retries work passed to `AuroraDSQLPool.transaction()`; it does not override `pool.query()`, so DDL and the separate tracking insert can fail immediately on `OC000`/`OC001`/`40001`. Please apply the adapter retry helper to each individual autocommit operation while preserving one DDL per transaction, and test a wrapped `40001` followed by success.

node/drizzle/src/session.ts
```diff
@@ -0,0 +1,161 @@
...
+      : this;
+
+    const tx = new AwsDsqlTransaction<TFullSchema, TSchema>(
+      this.dialect,
+      session,
+      this.dsqlSchema,
+    );
+
+    const configSql = buildTransactionConfigSQL(config);
+    await tx.execute(configSql ? sql`begin ${configSql}` : sql`begin`);
```
amaksimo: `BEGIN` is outside the `try/finally`. Once `connect()` has checked out a client, a `BEGIN` failure skips `release()` and leaks a pool slot. Also, if `ROLLBACK` fails below, that error replaces the original and the potentially broken client is released normally. #5538 keeps `BEGIN` inside the cleanup scope and preserves the original error; connector 0.1.9 discards the client after rollback failure. Please align these failure paths and add pooled-client tests for `BEGIN` and rollback failures.

node/drizzle/examples/veterinary-app/test/dsql-client.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    await expect(
+      db.transaction(async (tx) => {
+        // DSQL has no savepoints — failing fast on the client beats a
+        // cryptic server-side savepoint error.
+        await tx.transaction(async () => "unreachable");
+      }),
+    ).rejects.toThrow(/does not support nested transactions/);
+  });
+
+  test("concurrent transactionWithRetry contending on one row all commit", async () => {
```
amaksimo, line 262: This still does not guarantee or observe an OCC retry. The eight transactions can serialize naturally, and the final-value assertion passes even if the callback never reruns. [drizzle-orm#5538](https://github.com/drizzle-team/drizzle-orm/pull/5538) forces a conflict by committing a competing transaction after the first read and then asserts the callback attempt count; it also covers live exhaustion and verifies regular `transaction()` does not retry. Please mirror that deterministic pattern before resolving the earlier coverage concern.

amaksimo (commented):
Follow-up review after the latest changes.

praba2210:
Thanks for the thorough pass — one comment rather than 24 threads. All in `3c11466`: 131 unit tests, 15/15 live.

## Fixed

| Feedback | What changed |
|---|---|
| Don't default to `admin` | `user` is now required on `AwsDsqlOptions`, plus a runtime guard for JS callers. You were right this was ours — drizzle was the only adapter defaulting in *shipped* code. Dropping the `??` alone wasn't enough: `pg` falls back to `PGUSER`/the OS username. The example requires `CLUSTER_USER` too, matching prisma/sqlalchemy/hibernate. |
| Drop "DSQL does not support…" | Swept both READMEs, the nested-transaction error, the `migrate()` docblock, the changelog. Two kept: the AWS doc *title*, and a dsql-lint error about *its own* JSON version. |
| Wait on all async DDL | `ASYNC_DDL_RE` covers `CREATE [UNIQUE] INDEX ASYNC` and `ALTER TABLE ASYNC`; any `job_id` is awaited before recording, and a known async DDL returning a row without one fails. Test at `tests/migrator.test.ts:463`. |
| Autocommit ops bypass OCC retry | Correct, the connector doesn't cover `pool.query()`. All eight sites now use `withOccRetry`, pinned by a `test.each` injecting one wrapped `40001` each — including the migration's own DDL, which I'd missed because `CREATE TABLE "a"` doesn't match the tracking table's guarded `CREATE TABLE IF NOT EXISTS`. |
| "Safely re-run" oversells it | README documents the applied-but-untracked gap; `migrate()` names the statement to reconcile. |
| `BEGIN` outside `try/finally` | Moved inside. A failed `ROLLBACK` now preserves the original error and releases with `destroy: true` rather than returning a dirty connection, matching connector 0.1.9. Three pooled-client tests. |
| Contention test proves no retry | Agreed — eight transactions can serialize. Now #5538's forced conflict, asserting the retry itself: attempts > 1, `onRetry` fired, SQLSTATE `40001`, final value 101 not 1. Plus live exhaustion, and plain `transaction()` not retrying. |
| Test the dsql-lint output shape | The real gap — statuses and exit codes were covered, the shape guards weren't. Non-JSON stdout and an unknown `schema_version` now run end-to-end via a stub on `DSQL_LINT_PATH`. |
| Never pin dsql-lint | The caret was the bug: `^0.2.14` means `<0.3.0` on a 0.x package. Now `>=0.2.14 <1`, connector `>=0.1.9 <1`. Not `*`, since we hard-fail on an unknown `schema_version`. Happy to raise the ceiling once that's stable. |
| Missing from compat-canary | Added, mirroring `prisma`. |

## Not changed

| Feedback | Why |
|---|---|
| Hide stock `migrate()` | Not ours to hide — it's `drizzle-orm/node-postgres/migrator` and we never re-export it. It can't misbehave quietly either: its own tracking table is `id SERIAL PRIMARY KEY`, which DSQL rejects, so it dies before touching user DDL. Both do default to `drizzle.__drizzle_migrations` — say the word and I'll rename ours. |
| Move the example's `drizzle-orm`/`pg` into the adapter | Peer dependencies, so the app must install them. A second copy breaks subclassing (`NodePgSession`/`NodePgTransaction`/`PgDatabase` must come from your `drizzle-orm`) and `instanceof Pool`. `^0.45.2` is a range, not a pin. |
| Make drizzle-kit emit DSQL-safe SQL | No extension point — `dialect` is a closed union of six literals, so it means forking. And it would drift: DSQL's surface keeps growing, dsql-lint tracks it. |

## From running it live

- `pet_owner_id_idx` was recorded as applied while the index didn't exist — your bug in the wild, from the pre-fix migrator. Cleared the row; `migrate()` re-applied it, waited, index exists.
- The two live suites can't run concurrently: migrator DDL bumps the catalog, then the DML suite hits `OC001`. CI already runs them separately, `npm test` didn't, so jest pins `maxWorkers: 1`.

Retry ownership is intentional and now in the description: the adapter owns the loop so it can walk `DrizzleQueryError.cause`, which does bypass the connector's own OCC retry, so there are two implementations in the stack.

node/drizzle/examples/veterinary-app/src/veterinary-service.ts
```diff
@@ -0,0 +1,62 @@
...
+      .returning();
+    return row!;
+  }
+
+  async createSpecialty(name: string) {
+    const [row] = await this.db.insert(specialty).values({ name }).returning();
+    return row!;
+  }
+
+  async createVet(name: string, specialtyNames: string[]) {
```
amaksimo, line 41: The example demonstrates ORM relations but not application-level integrity. `createPet()` accepts an unchecked owner ID, and `createVet()` neither verifies specialty names nor makes the vet and join-row inserts atomic. Because these relationships are ORM-only, please validate referenced rows and perform each related operation in one `transactionWithRetry()`, with tests for invalid references and rollback. The Django pet-clinic example provides a repository-local pattern.

node/drizzle/examples/veterinary-app/README.md
```diff
@@ -0,0 +1,51 @@
...
+CREATE SCHEMA myschema;
+CREATE ROLE myuser WITH LOGIN;
+AWS IAM GRANT myuser TO 'arn:aws:iam::<account-id>:role/<your-iam-role>';
+GRANT USAGE, CREATE ON SCHEMA myschema TO myuser;
+GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA myschema TO myuser;
+```
+
+`CLUSTER_USER=myuser` makes the app connect with `search_path=myschema`. There is no default, so the app never lands on `admin` and the shared `public` schema by omission; `CLUSTER_USER=admin` still works if you want that for a first run.
+
+Note that `migrate()` creates its own tracking schema, so the role running migrations needs schema-creation privileges. A common split is to run `db:migrate` with an elevated identity during deployment and run the application itself as the scoped role.
```
amaksimo: The documented migration/runtime role split is not executable as written. `createDsqlDb()` sends `admin` to `public` and every non-admin role to `myschema`, so admin migrations create application tables outside the schema the app queries. A separate non-admin migrator has the other problem: the earlier `GRANT ON ALL TABLES` does not cover tables it creates later. Please document one complete setup, such as the same role, a post-migration grant, or default privileges for migrator-owned objects, and cover it with a scoped-role smoke test.

node/drizzle/src/retry.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+  return false;
+}
+
+/**
+ * Run `fn`, retrying on DSQL OCC conflicts with exponential backoff and equal
+ * jitter. Non-retryable errors propagate immediately; exhausting the retries
+ * throws {@link AwsDsqlRetryExhaustedError}.
+ */
+export async function withRetry<T>(
```
amaksimo, line 117: `withRetry()` is part of the package root export, but validation only occurs in `transactionWithRetry()`. Direct callers can pass `NaN`, fractional, or infinite `maxRetries`; the callback may not run, the original error can be lost, or retries can be unbounded. Please validate inside `withRetry()`, or stop exporting the helper. The primary `db.transactionWithRetry()` path is already safe; this issue is also present in upstream drizzle-team/drizzle-orm#5538.

amaksimo (commented):
One repository-fit item: the root README is unchanged, so Drizzle is absent from the Node.js adapter table, installation examples, and documentation links. Please add it before merge.

amaksimo (approved):
lgtm

PR #588, opened 2026-08-24
docs(django): clarify migration compatibility workflow

amaksimo:
## Summary

- keep Django's normal migration workflow and document `sqlmigrate | dsql-lint` only as an optional compatibility diagnostic
- narrow the known issue to unsupported `ALTER COLUMN TYPE` operations
- add a regression test for Django's inherited `RemoveField` SQL

## Audit context

Django already handles supported migration operations through its PostgreSQL schema editor, so the adapter does not need a table-recreation override. `dsql-lint 0.2.14` accepts the generated `DROP COLUMN ... CASCADE` SQL and reports remaining unsupported operations such as physical type changes.

## Testing

- unit suite exercised with Django 4.2, 5.0, 5.1, and 5.2: 96 passed and 8 subtests passed for each version
- `ruff check aurora_dsql_django/tests/unit/test_wrapper.py`
- generated `DROP COLUMN ... CASCADE` SQL checked with `dsql-lint 0.2.14`


praba2210:
LGTM!

praba2210 (approved):

PR #597, opened 2026-08-26
Update changelog for node/drizzle/v0.1.0

praba2210:
Generated by the `update-changelog` workflow for `node/drizzle/v0.1.0`, which
rewrites the hand-written `[Unreleased]` section into the per-version format
used by the other adapters.

Opened manually because the workflow's own final step fails with
`Resource not accessible by personal access token (createPullRequest)` —
`BOT_GITHUB_TOKEN` can push the branch but cannot open a PR. Pre-existing:
the same step failed on the previous run in February, which is why recent
releases carry no changelog entries. Worth fixing separately.

*Issue #, if available:*
N/A


node/drizzle/CHANGELOG.md
```diff
@@ -1,36 +1,36 @@
...
+
+See README for full documentation.
+
+## Contributors
+
+- Add Aurora DSQL adapter for Drizzle ORM by [@praba2210](https://github.com/praba2210) in [#563](https://github.com/awslabs/aurora-dsql-orms/pull/563)
+- Scope the Drizzle release gate to its own checks by [@praba2210](https://github.com/praba2210) in [#595](https://github.com/awslabs/aurora-dsql-orms/pull/595)
+- Allow a token for a package's first npm publish by [@praba2210](https://github.com/praba2210) in [#596](https://github.com/awslabs/aurora-dsql-orms/pull/596)
+
+Thanks to [@amaksimo](https://github.com/amaksimo) for the review.
```
anwesham-lab: I think we usually want this in a what's changed structure? Refer to one of the older changelog structures like the python sqlalchemy example.

praba2210: Updated the format, thanks!

anwesham-lab (approved):

PR #598, opened 2026-08-27
Foreign key feature release updates

davidrz15:
Part of the rollout along with: https://github.com/awslabs/aurora-dsql-tools/pull/142

PR #601, opened 2026-08-28
Rely on native foreign keys in ORM samples

amaksimo:
## Summary

- rely on native foreign-key enforcement in the Drizzle sample instead of race-prone application existence checks
- preserve foreign-key violation and transaction rollback coverage
- correct the Hibernate 1.0 release history and scope the Tortoise foreign-key backfill guidance to databases created with 0.2.0 or earlier
- refresh the branch onto current `main` and drop stale Prisma and Django changes; Prisma validation and packaging will be handled separately

## Verification

- Drizzle: format, build, 134 package tests
- Hibernate dialect: Gradle build
- Tortoise ORM: `uv lock --check`
- Drizzle foreign-key migration: clean with `dsql-lint`
- `git diff origin/main...HEAD --check`

Live DSQL checks remain in CI.


java/hibernate/examples/pet-clinic-app/build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
   id 'java'
-  id 'org.springframework.boot' version '4.1.1'
+  id 'org.springframework.boot' version '4.1.0'
```
anwesham-lab: why downgrade version?

anwesham-lab:
### Code review

Net −611/+113 with the FK enforcement moved into the database is the right direction, and most of it holds up: both rewritten migrations apply verbatim to a live Aurora DSQL cluster with **zero** `dsql-lint` diagnostics, all three FKs come back `convalidated = t` (so no `NOT VALID` / async-validation step is needed for fresh installs, and deleting `validate-foreign-keys.ts` is correct for that path), every declared referential action behaves as written, and the `down.sql` reorder is a **required** fix — the old parent-first order now fails with `2BP01`.

Two blocking issues, both the same shape: a database created by an **already-released** version silently ends up with no foreign keys, and this PR removes every mechanism that would have caught it.

---

#### 1. Rewriting the released initial migrations leaves existing databases permanently FK-less, silently

`0_init` shipped in [`node/prisma/v0.1.0`](https://github.com/awslabs/aurora-dsql-orms/releases/tag/node/prisma/v0.1.0) (2026-02-03) and `0000_init` in [`node/drizzle/v0.1.0`](https://github.com/awslabs/aurora-dsql-orms/releases/tag/node/drizzle/v0.1.0) (2026-08-25), both **without** foreign keys. Folding the FKs into those files means they are never applied to a database that already recorded them.

Reproduced end-to-end against a live DSQL cluster:

```
# Prisma: apply the released FK-less 0_init, then upgrade to this PR's version
$ npx prisma migrate deploy   # released 0_init
All migrations have been successfully applied.        -> pg_constraint contype='f': 0 rows

$ npx prisma migrate deploy   # this PR's 0_init, FKs inlined
1 migration found in prisma/migrations
No pending migrations to apply.                       -> pg_constraint contype='f': 0 rows
```
No checksum warning, no error, exit 0. Prisma's own guidance is [do not edit applied migrations](https://www.prisma.io/docs/orm/prisma-migrate/understanding-prisma-migrate/migration-histories).

```
# Drizzle: same experiment via the adapter's own migrate()
$ npm run db:migrate          # this PR's 0000_init
Warning: statement 1 in migration 20260812032724 has changed since it was applied ... It will be skipped.
Warning: statement 3 in migration 20260812032724 has changed since it was applied ... It will be skipped.
Warning: statement 4 in migration 20260812032724 has changed since it was applied ... It will be skipped.
Applied 0 statement(s) across 0 migration(s).         -> pg_constraint contype='f': 0 rows
```

That warn-and-skip path is the one added at review request in #563 (`migration.hash` → stable journal identity + statement index); it is now firing against this change. Statements 3 and 4 also swap identity because `specialty_to_vet` moved after `vet`.

The consequence, also run on the cluster — on that FK-less database the post-PR sample commits an orphan with no error:

```sql
INSERT INTO "pet" ("id","name","birthDate","ownerId")
VALUES (gen_random_uuid(),'SilentOrphan','2020-01-01','00000000-0000-4000-8000-000000000000');
-- succeeds; row is committed
```

Nothing is left to detect it, because the same PR removes both guards:
- the owner/specialty existence checks in `createPet`/`createVet` — https://github.com/awslabs/aurora-dsql-orms/blob/72efb4fd4e44707f1999c65626d124f051fed70c/node/drizzle/examples/veterinary-app/src/veterinary-service.ts#L15-L23
- `validate-foreign-keys.ts`, which threw ``Foreign key ${foreignKey.name} is missing; run Prisma migrations first`` and was wired into `prisma:migrate-up` — https://github.com/awslabs/aurora-dsql-orms/blob/043b210ee205e337fec9b1dca5ed838f3910f653/node/prisma/examples/veterinary-app/src/validate-foreign-keys.ts#L33-L36

Keeping the inline FKs is right for fresh installs. Please also keep a forward path for databases already at `0_init`/`0000` — restoring the deleted `1_enable_foreign_keys` / `0001` (idempotent `ADD CONSTRAINT ... NOT VALID` + `ALTER TABLE ASYNC ... VALIDATE CONSTRAINT`) is the cheapest option, since that code already existed and was tested. Failing that, keep `validate-foreign-keys.ts` in `prisma:migrate-up` so the state fails loudly instead of silently.

https://github.com/awslabs/aurora-dsql-orms/blob/72efb4fd4e44707f1999c65626d124f051fed70c/node/prisma/examples/veterinary-app/prisma/migrations/0_init/migration.sql#L30-L35

https://github.com/awslabs/aurora-dsql-orms/blob/72efb4fd4e44707f1999c65626d124f051fed70c/node/drizzle/examples/veterinary-app/drizzle/0000_init.sql#L11-L16

#### 2. Tortoise: the only documented upgrade path for FK-less databases was deleted, and it is still needed

`python/tortoise-orm/v0.2.0` (released 2026-07-13) suppressed every FK constraint:

```python
def _create_fk_string(self, ...) -> str:
    """Return empty string since DSQL doesn't support foreign key constraints.
    ...
    """
    return ""
```

That override was only removed in `23da334` (#598), which is **unreleased** (`git tag --contains 23da334` is empty). So databases created by shipped v0.1.0–v0.2.0 have no FK constraints today, and as the deleted section itself explained, *"Aerich model-state comparison won't generate this migration because the relationship was already present in the model state"* — so upgrading will not add them. Nothing in this PR makes that guidance false, and no replacement survives anywhere (`grep -rn "existing database" --include="*.md"` finds only the unrelated `--from-config-datasource` note).

https://github.com/awslabs/aurora-dsql-orms/blob/043b210ee205e337fec9b1dca5ed838f3910f653/python/tortoise-orm/docs/ADAPTER_BEHAVIOR.md#L77-L96

---

#### Non-blocking, in priority order

**a. The deleted cascade row-limit warning describes an unrecoverable failure that the Prisma sample still has.** `_SpecialtyToVet_A_fkey` is `ON DELETE CASCADE`. On a live cluster with 3,500 vets holding one specialty:

```
DELETE FROM "specialty" WHERE "name"='bulk';
ERROR:  54000: transaction row limit exceeded
CONTEXT:  SQL statement "DELETE FROM ONLY "public"."_SpecialtyToVet" WHERE $1 = "A""
```

The parent then cannot be deleted at all without manually batch-deleting children first. The removed paragraph said exactly this, and the [DSQL docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/create-table-syntax-support.html#create-table-fk-actions) say it too ("Prefer `NO ACTION` or `RESTRICT` ... where child-row cardinality is unbounded"; [3,000-row limit](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/CHAP_quotas.html#SECTION_database-limits)). Dropping the `relationMode` framing was right; consider keeping one sentence plus a link, rather than dropping the caveat with it. Note the Drizzle sample uses the doc-preferred `RESTRICT` for the same logical model.

https://github.com/awslabs/aurora-dsql-orms/blob/043b210ee205e337fec9b1dca5ed838f3910f653/node/prisma/README.md#L155-L160

**b. A released changelog entry was rewritten.** `[1.0.0] - 2026-01-29` genuinely had no FK support (`NO_FK_SUPPORT_EXPORTER` survived until #598), and the same file still tells Hibernate 6.6 / Spring Boot 3.x users to stay on 1.0.x — so those readers now get a 1.0.0 entry with the limitation erased. Deleting the `Unreleased` section is fine; this line is history.

https://github.com/awslabs/aurora-dsql-orms/blob/72efb4fd4e44707f1999c65626d124f051fed70c/java/hibernate/CHANGELOG.md#L29-L31

**c. `createPet` lost its `transactionWithRetry` wrapper**, so a correctly-classified OCC conflict is now propagated to the caller instead of retried. Removing the pre-check is correct; dropping the retry is separable.

Worth stating what is *not* wrong here, since it bounds the fix. DSQL classifies these cleanly in both directions, verified on a live cluster:

| Scenario | Result |
|---|---|
| Genuinely missing parent | `23503` at **statement** time |
| Child insert loses to a concurrent parent delete | `40001` / `OC000` at **commit** |
| Parent delete loses to a concurrent child insert | `40001` / `OC000` at **commit**, no dangling child |

So an OCC conflict never masquerades as a functional FK violation, and since the retry predicate is exactly `{OC000, OC001, 40001}`, the deterministic `23503` in `createVet` is never retried — the new tests asserting `23503` are correct and cannot flake into a retry-exhausted error. The only gap is that `createPet` now lets the `40001` escape raw, where main retried it, and `node/drizzle/README.md` still says to "use transaction retry handling because foreign-key conflicts can surface as serialization failures". Counterpoint: `createOwner` and `createSpecialty` were already bare single-statement inserts on main, so this does match existing house style — your call.

https://github.com/awslabs/aurora-dsql-orms/blob/72efb4fd4e44707f1999c65626d124f051fed70c/node/drizzle/examples/veterinary-app/src/veterinary-service.ts#L15-L23

**d. `drizzle/0000_init.sql` is no longer reproducible by the documented tooling.** `drizzle-kit generate` on the unchanged `src/schema.ts` emits tables alphabetically with FKs as separate `ALTER TABLE ... ADD CONSTRAINT` (which `dsql-lint --fix` then marks `NOT VALID`), not inline, and not with `specialty_to_vet` after `vet`. `db:generate` is a no-op today so nothing drifts now, but the README's rewrite list ("`CREATE INDEX ASYNC`, `USING btree` removed") omits the two larger manual edits.

https://github.com/awslabs/aurora-dsql-orms/blob/72efb4fd4e44707f1999c65626d124f051fed70c/node/drizzle/examples/veterinary-app/README.md#L7-L9

---

#### FYI, not attributable to this PR

`gate` is red on `72efb4f`: `java-hibernate / Test Sample (DSQL + gradle)` fails 4 of 44 (`ClinicServiceTests.shouldAddNewVisitForPet` and 3 siblings, all `ConstraintViolationException`). Identical failures on `main` since #598, which merged with the gate bypassed. The `4.1.1 → 4.1.0` change here is not the cause — the Maven sample passes at 4.1.0. Worth a separate fix so `main` goes green.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anwesham-lab:
I think the blocking findings are a matter of ordering. Do we need to vend the upstream changes in advance of this going in?

java/hibernate/examples/pet-clinic-app/build.gradle
amaksimo: Reverted. The downgrade was unrelated to native foreign keys, and the Gradle DSQL failures also occurred with 4.1.0.

anwesham-lab (approved):
we should validate why checks are failing and how to fix, but lgtm

Worth noting, the Maven DSQL job is not testing the dialect built by this PR. That might be part of the problem. The dialect artifact is version `2.1.0`, but the Maven sample still depends on `2.0.0`; Gradle loads the downloaded `2.1.0` JAR directly so the failure could be relevant. 

This makes Maven’s green result misleading given the corresponding Gradle tests fail. We might need to align Maven with the build artifact, probably by injecting the downloaded JAR version to rerun the jobs again. If it also fails, I'd be more concerned about the Gradle check as well.

